PR #4, opened 2019-03-26
Add a `<Header />`, `<AppTitle />` and `<AI2Logo />` component.

codeviking:
This adds components that support rendering a standard application
header.

Part of this required modifying the default theme. I added colors
per the AI2 website and deleted the `testColor` mechanism we
were using to prove out that approach.


codeviking:
@jonborchardt Here you go!

src/lib/components/Header.tsx
```diff
@@ -0,0 +1,41 @@
+import * as React from 'react';
+import { AppBar, Grid } from '@material-ui/core'
+import styled from 'styled-components';
+
+interface Props {
+    position?: 'fixed' | 'absolute' | 'sticky' | 'static' | 'relative',
```
jonborchardt: move to types?
type Position = 'fixed' | 'absolute' | 'sticky' | 'static' | 'relative';
then
interface Props {
    position?: Position

src/lib/components/Header.tsx
```diff
@@ -0,0 +1,41 @@
...
+ * applications.
+ */
+export const Header = (props: Props) => {
+    const { children, position } = props;
+    // Make sure the children are an array, as we allow them to specify
+    // a single descendent.
+    const childArray = Array.isArray(children) ? children : [ children ];
+    return (
+        <StyledAppBar position={position}>
+            <Grid container spacing={24} alignItems="center" wrap="wrap">
```
jonborchardt: we could expose more of these... but not needed

src/lib/components/Header.tsx
```diff
@@ -0,0 +1,41 @@
...
+}
+
+const StyledAppBar = styled(AppBar)`
+    border-top: 4px solid ${props => props.theme.palette.highlight.main};
+    padding: 1rem 2rem;
+    && {
+        background: #fff;
+    }
+`
+
```
jonborchardt: remove extra lines

src/lib/components/Header.tsx
codeviking: 👍 I wonder if I can extract the type from their definition. I'll dig in and see what I can do.

src/lib/components/Header.tsx
codeviking: You mean as as a separate component, i.e. a `<FlexRow />` or something like that?

src/lib/components/Header.tsx
codeviking: 👍

src/lib/components/Header.tsx
jonborchardt: no, i mean we could allow user to pass in more props.... wrap, spacing, and just set these as defaults

codeviking: Ah! Totally. I figured I'd leave those out until we need them -- as maybe people won't need them at all!

PR #6, opened 2019-03-26
Add a full page example and supporting components.

codeviking:
This:

* Adds a full page example, and the ability to route there (via
  `react-router`).  Going to `/` (in the demo app) routes to the
  original UI (the "component gallery") while `/full-page` goes
  to the full page example.
* Adds a `<MaxWidthCenteredContent />` and `<Main />` elements
  which allow us to easily compose layouts that don't grow beyond
  a certain width and are horizontally centered.
* Changes the default font size to `16px`, which is usually the
  case in our applications.


src/examples/FullPage.tsx
```diff
@@ -0,0 +1,21 @@
+import * as React from 'react';
+import { Typography } from '@material-ui/core';
+
+import { ThemeProvider } from '../lib';
+import { Header, AI2Logo, AppTitle, Main } from '../lib/components';
+
+const FullPage = () => (
+    <ThemeProvider>
```
jonborchardt: i kind of expect the themeporovider to live in index.tsx

src/examples/FullPage.tsx
```diff
@@ -0,0 +1,21 @@
...
+        <Header>
+            <AI2Logo />
+            <AppTitle>Full Page Example</AppTitle>
+        </Header>
+        <Main>
+            <Typography variant="body1">
```
jonborchardt: consider converting to a styled component
we could do them all
const Body1 = styled(Typography).attrs(p=>{variant: "body1"})``;

src/examples/FullPage.tsx
codeviking: 👍 I only did it here as I was following the convention in `App.tsx`, but I understand why it should be there and not here.

Moving!

src/examples/FullPage.tsx
codeviking: Interesting. I like the idea, and it follows suit with what I've been doing elsewhere.

I think I'll just call it `<BodyText>` for now.  The motivation / idea being that we're trying to provide a simple, opinionated subset of controls to our users -- if they want to break out of the mold and use react-material they're perfectly able to do so, but we make it easy for them to not do so.

PR #33, opened 2019-06-20
Jon/text links

jonborchardt:
see 2 commits for a clearer set of cahnges


ui/src/demo/pages/About/About.tsx
```diff
@@ -2,7 +2,7 @@ import * as React from 'react';
 import styled from 'styled-components';
 import { RouteComponentProps } from 'react-router';
 
-import { Paragraph, ExternalLink, Body, H2 } from '../../../lib/components';
+import { Paragraph, TextLink, Body, H2 } from '../../../lib/components';
```
codeviking: What's a "TextLink" vs a "Link"?

Also, why not just apply these styles globally to `a[href]`, thereby allowing users to just use good ole' `<a href="..">Link</a>`.  This way there's no need to import a component just to add a link -- users just use HTML like they'd expect.

I get having components for buttons and other stuff -- but where possible I'd encourage us to fallback to default HTML tags where possible to reduce cognitive overhead for our users.

codeviking: There is, of note, an exception to this when using react router, as there we should be using a `NavLink`.  The `NavLink` ensures the transition is client-side (without going back to the server) whereas a traditional link `<a href />` will cause the page to reload (which is pointless when using react-router, as we're just adding extra network overhead).

jonborchardt: we do update the base html elements
the css overrides make:
```
a {
  color: #2496CB;
  text-decoration: none;
  background-color: transparent;
  outline: none;
  cursor: pointer;
  transition: color 0.3s;
  -webkit-text-decoration-skip: objects;
}
a:hover {
  color: #2496CB;
}
a:active {
  color: #2496CB;
}
a:active,
a:hover {
  text-decoration: underline;
  outline: 0;
}
a[disabled] {
  color: rgba(0, 0, 0, 0.25);
  cursor: not-allowed;
  pointer-events: none;
}
```

which is fine....
but we cannot do dark this way

ui/src/demo/pages/Design/Colors.tsx
```diff
@@ -2,15 +2,116 @@ import * as React from 'react';
...
+        colors.N10
+    ]
+};
+
 export class Colors extends React.PureComponent<RouteComponentProps> {
     render() {
         return (
             <React.Fragment>
-                <PageHeader>Colors</PageHeader>
+                <PageTitle>Colors</PageTitle>
```
codeviking: What differentiates a `<PageTitle />` from a good ole' `<h1 />`?

jonborchardt: margin

i have an open q to nick on if our headings should use std margins

ui/src/demo/pages/Design/TextLinks.tsx
```diff
@@ -0,0 +1,77 @@
...
+    display: inline-block;
+    padding-right: ${props => `${convertPixelsToRem(32)}rem`};
+    margin-top: ${props => `${convertPixelsToRem(16)}rem`};
+`;
+
+const DarkBody = styled(LightBody)`
+    background: ${({theme}) => theme.colors.B9.value};
+    padding: ${props => `${convertPixelsToRem(16)}rem`};
+    border-radius: ${({theme}) => `${theme.shape.borderRadius}px`};
+    color: ${({theme}) => theme.palette.white};
```
codeviking: I'm concerned about all of these custom padding and margin values we're spreading around the codebase. It seems like it's going to be tricky / a lot of work to update them.

Even if Nick hasn't specified them yet, I'd encourage us to add them to the theme.  We can always update them to be the values he'd expect.  Leaving that for later is just creating extra work, as there will be more values to update.

I'd take that on now.

ui/src/demo/pages/Design/Colors.tsx
jonborchardt:
```
margin-top and bottom are (by default) different for h1-h6
your spec does not specify margins… is this because
1- ai2 headlines should not have a top or bottom margin
2- you want the default browser margin for headlines?
```

ui/src/demo/pages/Design/TextLinks.tsx
jonborchardt: ok, ill try to take a pass
im concerned that its taking longer than expected to get to grover, but i agree this is worth it

ui/src/lib/components/textLink.tsx
```diff
@@ -0,0 +1,19 @@
+import * as React from 'react';
+import styled from 'styled-components';
+
+
+export const TextLink = styled.a<{active: boolean, dark: boolean}>`
```
codeviking: I actually liked the `ExternalLink` name better (sorry for back and forth on this), as we need to differentiate between:

* Links that are to different parts of the SPA, which should use react router's `NavLink` or something that extends from it.
* Links to other websites.

Maybe we could make a single component that somehow handles that elegantly? If we detect that the link is offsite (i.e. to a different origin), we use a `<a href="" />`, if not we use a `<NavLink />`, but we make sure they look the same?

jonborchardt: ill make sure they look the same...
but for now, there will be 2... sound ok?

ui/src/lib/components/textLink.tsx
```diff
@@ -0,0 +1,19 @@
...
+        : (dark ? theme.textLink.contrastColor.value : theme.textLink.color.value)
+    )};
+    decoration: ${({theme}) => theme.textLink.decoration};
+
+    &:hover {
+        color: ${({ theme, dark }) => (
+            dark ? theme.textLink.contrastHoverColor.value : theme.textLink.hoverColor.value
+        )};
+        decoration: ${({theme}) => theme.textLink.hoverDecoration};
+    }
```
codeviking: Rather than inlining all of these, could we just do something like:

```
const colors = dark ? { active: theme.textLink.contrastActiveColor.value, ... } : {}
```

Then in the style declarations just reference `colors`?

I feel like it'd make it a bit easier to follow.

Also, I'm concerned about the extra overhead of referencing `.value` for every color. What's the value of having a separate member for this? Would a user ever use `displayName`? It feels like a bit of an awkward abstraction, and it's different from what I'm familiar with from other theming libraries where I just reference `theme.color.blah`.  I might catch this one earlier rather than later and simplify it so colors are simply the hex values they refer to, not objects, in the interest of establishing a simpler API.

ui/src/lib/theme/generateOverrides.ts
```diff
@@ -123,11 +125,11 @@ const template =`
...
-    @link-color: ${colors.B7.value};
-    @link-hover-color: @link-color;
-    @link-active-color: @link-color;
-    @link-decoration: none;
-    @link-hover-decoration: underline;
+    @link-color: ${textLink.color.value};
+    @link-hover-color: ${textLink.hoverColor.value};
+    @link-active-color: ${textLink.activeColor.value};
+    @link-decoration: ${textLink.decoration};
+    @link-hover-decoration: ${textLink.hoverDecoration};
```
codeviking: Why do we need to set these in the ant theme and our components? Doesn't this cause the theme'd css to style `a[href]` tags by default, without needing to use an explicit components?

A clarifying question / way to investigate this further -- when you inspect the elements, do you see the styles applied here declared twice, one from the ant definition and another from the style injected by the component? If so we definitely don't want that, as we're making the browser do 2x the work (and including 2x the css).

ui/src/lib/theme/textLink.ts
```diff
@@ -0,0 +1,15 @@
...
+
+const color = colors.B7;
+const contrastColor = colors.B4;
+
+export const textLink = {
+    color: color,
+    hoverColor: color,
+    activeColor: color,
+    decoration: 'none',
+    hoverDecoration: 'underline',
```
codeviking: Personally I'm a fan of the underline just being present all the time. It helps me differentiate links from colored text (as a user who isn't so adept at detecting differences derived from color only).

In other words, I think displaying an underline on links all the time makes them easier for colorblind users to see (and thereby more accessible).  Ultimately I defer to @Pelagic-sky's preference here, but wanted to share the thought.

ui/src/demo/pages/About/About.tsx
codeviking: What's dark for? Do we have a dark theme and a light theme now? Are we sure we want that?

ui/src/demo/pages/Design/Colors.tsx
codeviking: 👍 Sounds like it's being driven by the design spec, which is the right call.

My hope would be is it's just the styles attached to a `<h1 />`, so we can reduce what users have to think about.

ui/src/lib/components/textLink.tsx
codeviking: Sure, let's just make sure they're named appropriately.  Right now I'd be confused as to whether I should use a `TextLink` for links to other pages on the same site (which I shouldn't, I should use a `NavLink`).  So maybe something like `NavLink` or `SiteNavLink` and `ExtLink` or `ExternalLink` or `OffsiteLink`?

ui/src/lib/components/textLink.tsx
jonborchardt: for the first part.. not sure what you mean... 

for the second part, i spent a lot of time going back and forth on the colors stuff...
i landed on this, because i needed the name with the color for display in the styleguide
so we can do stuff like ``` <li>Basic text links are defined with the color {color.displayName}.</li>```
however, i can revisit
the alternative is to add another layer here
to pull the theme from the 'styleguide info'
```
export const textLink = {
    color: styleGuide.textLink.color.value,
    hoverColor: styleGuide.textLink.hoverColor.value,
    //etc...
};
```
or to add the color names as a sep entity 'next' to the color
```
export const textLink = {
    color: color,
    colorName: string,
    hoverColor: color,
    hoverColorName: string,
    //etc...
};
```

ui/src/lib/theme/generateOverrides.ts
jonborchardt: these do 2 things
1- update the base css (as i pasted in a previous discussion)
2- update ant to use them inside their general components

OUR components will also use these values.... 

i see it like this
we have theme files as our root (basically json)
  those are used to override ant AND set direct values to css (ie, h1s)
  those are also used in any styled components we write

ui/src/lib/theme/textLink.ts
jonborchardt: up to design IMO

ui/src/demo/pages/About/About.tsx
jonborchardt: when the user knows the link will be shown on dark background.... they need to tell the link to use a different set of colors

jonborchardt: dark is a bad name... ill change regardless

ui/src/demo/pages/Design/Colors.tsx
jonborchardt: agreed, ill nudge in that direction

ui/src/lib/components/textLink.tsx
jonborchardt: agreed, ill have you rereview before merging

ui/src/lib/theme/textLink.ts
Pelagic-sky: IMG_6725
IMG_6724
Testing the colors against the four main types of colorblindness, I can still distinguish text against a link. My personal preference is to not overdo text links with underlines as it takes away from the reading experience by being too intrusive. 
If want to get fancy, TechCrunch.com has a visually elegant solution that lets an underline exist, however, it is a lighter grey when inactive and fades into the color on hover. I'm attaching a screenshot for reference, but you can view the effect by looking at an article. https://techcrunch.com/2019/06/20/is-seed-investing-still-a-local-business/
Screen Shot 2019-06-20 at 2 06 48 PM

ui/src/demo/pages/About/About.tsx
codeviking: Ah, `inverse`, or something like that.

Interesting. I suppose I'm curious how often that happens.  I haven't had it happen in developing a demo yet.

That said if it's in the style guide we should support it.  Maybe we just called it `inverse`. 

Sounds like we only need to apply styles if it's inversed, and if not we should just use the link styles applied by Ant's css by default, right?

ui/src/lib/theme/generateOverrides.ts
codeviking: Sure, let's just make sure we don't overwrite the styles with styles that do the same thing (i.e. reapply the default styles), as that's 2x the payload size, specificity, and confusing.

I think I noted this above where I implied that by default, the link can just use the styles originating from the CSS -- the inverse treatment is the only one that necessitates any JS based styles.

Does that make sense?

ui/src/lib/components/textLink.tsx
codeviking: Ah, gotcha.

I lean towards keeping the API that users use really simple.  The need to label the colors (give them a name) seems like a concern that's specific to the Varnish demo.

Could we make a mapping that converts a color key to it's name, so that the actual definition is separate from the names?

In practice it'd look something like:
```
const colorNamesByKey = { N6: "Neutral 6" }
const color = { N6: '#fff' };


Object.keys(color).map(key => ({ displayName: colorNamesByKey(key), value: color[key] })
```

Let me know what you think.

ui/src/demo/pages/About/About.tsx
jonborchardt: correct.... i think

ui/src/demo/pages/Design/Colors.tsx
jonborchardt: i have removed the styled component versions of Hs

ui/src/lib/utils/base.ts
```diff
@@ -44,11 +19,19 @@ function getRootFontSize(dynamic = false): number {
...
 }
 
 // convert px to rems using document font size
-export function convertPixelsToRem(px: number): number {
-    return px / getRootFontSize();
+export function convertPixelsToRemStr(px: number, dynamic = false): string {
+    return `${px / getRootFontSize(dynamic)}rem`;
+}
+export function convertPixelsStrToRemStr(pxStr: string, dynamic = false): string {
+    const px = parseFloat(pxStr.slice(0, -2));
```
codeviking: I don't think you need the slice here, as `parseFloat` just ignores the unit:

```
> parseFloat("30.5rem")
30.5
```

ui/src/demo/pages/Design/TextLinks.tsx
jonborchardt: done

ui/tsconfig.json
```diff
@@ -26,7 +26,7 @@
   },
   "include": [
     "./src/react-app-env.d.ts",
-    "./src/lib/**/*"
+    "./src/lib/**/*", "src/demo/AppRoute.ts"
   ],
```
codeviking: I'm confused why we need the include explicitly. I haven't needed this in other projects (I just pointed at a main file to compile and it went from there).

ui/src/lib/utils/base.ts
```diff
@@ -44,11 +19,19 @@ function getRootFontSize(dynamic = false): number {
 }
 
 // convert rems to px using document font size
-export function convertRemToPixels(rem: number, dynamic = false): number {
-    return rem * getRootFontSize();
+export function convertRemToPxStr(rem: number, dynamic = false): string {
+    return `${rem * getRootFontSize(dynamic)}px`;
+}
+export function convertRemStrToPxStr(remStr: string, dynamic = false): string {
+    const rem = parseFloat(remStr.slice(0, -3));
```
codeviking: Likely can drop the `slice` here too.

ui/src/lib/components/textLink.tsx
jonborchardt: ahh... i did something.... but i like what you suggest... ill think about it...

ui/src/lib/theme/generateOverrides.ts
```diff
@@ -82,52 +88,58 @@ const template =`
...
+
+    // vertical paddings
+    @padding-lg: ${convertRemStrToPxStr(spacing.lg)}; // containers
+    @padding-md: ${convertRemStrToPxStr(spacing.md)}; // small containers and buttons
+    @padding-sm: ${convertRemStrToPxStr(spacing.sm)}; // Form controls and items
+    @padding-xs: ${convertRemStrToPxStr(spacing.xs)}; // small items
```
codeviking: Why are we converting these to pixels? Why not just use rems?

ui/src/lib/theme/generateOverrides.ts
jonborchardt: sure... lets go over in person, because i think we have to do what im doing here....

ui/src/lib/theme/textLink.ts
jonborchardt: @Pelagic-sky this is your call IMO
soinds like i should leave it as is

ui/src/lib/utils/base.ts
jonborchardt: really... cool

ui/src/lib/theme/textLink.ts
codeviking: Sounds good, thx.

ui/tsconfig.json
jonborchardt: =ha, looks like a paste error

ui/src/lib/utils/base.ts
```diff
@@ -44,11 +19,19 @@ function getRootFontSize(dynamic = false): number {
...
+    return convertRemToPxStr(rem, dynamic);
 }
 
 // convert px to rems using document font size
-export function convertPixelsToRem(px: number): number {
-    return px / getRootFontSize();
+export function convertPixelsToRemStr(px: number, dynamic = false): string {
+    return `${px / getRootFontSize(dynamic)}rem`;
+}
```
codeviking: Is there a `convertPixelsToRem`? I'd drop the `Str` as the type is provided by TypeScript.

ui/src/lib/theme/generateOverrides.ts
jonborchardt: good point, there WAS a reason, but not now

ui/src/lib/utils/base.ts
jonborchardt: there was... now there is not ;-P

ui/src/lib/components/textLink.tsx
```diff
@@ -0,0 +1,19 @@
+import * as React from 'react';
+import styled from 'styled-components';
+
+
+export const TextLink = styled.a<{active: boolean, dark: boolean}>`
+    color: ${({ theme, active, dark }) => (
+        active
+        ? (dark ? theme.textLink.contrastActiveColor.value : theme.textLink.activeColor.value)
+        : (dark ? theme.textLink.contrastColor.value : theme.textLink.color.value)
```
codeviking: Should we rename `dark` to `inverse`?

It might be the right way to go -- that way if a theme is dark by default, inverse should actually be dark colors (as the inverse of the theme's standard background will be light)...

ui/src/lib/theme/colors.ts
```diff
@@ -1,208 +1,165 @@
...
+    N1: {displayName: 'N1', value: color.N1 },
+    N2: {displayName: 'N2', value: color.N2 },
+    N3: {displayName: 'N3', value: color.N3 },
+    N4: {displayName: 'N4', value: color.N4 },
+    N5: {displayName: 'N5', value: color.N5 },
+    N6: {displayName: 'N6 (Base)', value: color.N6 },
+    N7: {displayName: 'N7', value: color.N7 },
+    N8: {displayName: 'N8', value: color.N8 },
+    N9: {displayName: 'N9', value: color.N9 },
+    N10: {displayName: 'N10', value: color.N10 }
```
codeviking: Was it not possible to make this dynamic, i.e. only store `displayName` and build up the map programmatically?

ui/src/lib/components/textLink.tsx
jonborchardt: yes

ui/src/lib/theme/colors.ts
jonborchardt: yes, saw your suggestion after coded it up.... ill relook

ui/src/lib/theme/DefaultVarnishTheme.ts
```diff
@@ -1,26 +1,31 @@
...
-      primary: colors.B9,
-      primary2: colors.B8,
-      primary3: colors.B6,
-      primary4: colors.O6,
-      error: colors.R6
+      primary: colorInfo.B9,
+      primary2: colorInfo.B8,
+      primary3: colorInfo.B6,
+      primary4: colorInfo.O6,
+      error: colorInfo.R6
```
codeviking: I have a less crazy idea we chatted about in person:

```
class Color {
  constructor(value, displayName) {
    this.value = value;
    this.displayName = displayName;
  }
  toString() {
    return this.value;
  }
}

const blue = new Color('#fff', 'blue');

console.log(`${blue}`);
console.log(blue.displayName);
console.log('' + blue);
```

ui/src/lib/theme/generateOverrides.ts
codeviking: What I mean is that in styled components we shouldn't emit a style that we've already emitted via less / the default theme.

jonborchardt:
merging, will fix the links in sep pr

PR #34, opened 2019-06-24
add buttons to styleguide

jonborchardt:
and random fixups


ui/src/lib/components/button.tsx
```diff
@@ -0,0 +1,75 @@
...
+                ? theme.button[type].border || theme.button.default.border
+                : theme.button[type].contrastBorder || theme.button.default.contrastBorder || theme.button[type].border || theme.button.default.border
+        }};
+        color: ${({theme, type, contrast}) => {
+            return !contrast
+                ? theme.button[type].color || theme.button.default.color
+                : theme.button[type].contrastColor || theme.button.default.contrastColor || theme.button[type].color || theme.button.default.color
+        }};
+        font-weight: ${({theme, type}) => theme.button[type].fontWeight || theme.button.default.fontWeight};
+        height: ${({theme, type}) => theme.button[type].height || theme.button.default.height};
```
codeviking: We should avoid setting `height` so that things properly grow to encompass the content. Instead we should use `padding`, `font-size` and `line-height`.  This will equate to more flexible buttons.

ui/src/lib/components/button.tsx
```diff
@@ -0,0 +1,75 @@
+import * as React from 'react';
+import styled from 'styled-components';
+import { Button as AntButton } from 'antd';
+
+export type ButtonTypes = 'primary' | 'default' | 'link' | 'marketing';
+
+// todo: consider converting to use https://www.npmjs.com/package/styled-components-modifiers
```
codeviking: Seems fine -- I'd explore it as an alternative.

ui/src/lib/components/button.tsx
```diff
@@ -0,0 +1,75 @@
...
+            color: ${({theme, type, contrast}) => {
+                return !contrast
+                    ? theme.button[type].disabled.color || theme.button.default.disabled.color
+                    : theme.button[type].disabled.contrastColor || theme.button.default.disabled.contrastColor || theme.button[type].disabled.color || theme.button.default.disabled.color
+            }};
+            opacity: ${({theme, type}) => theme.button[type].disabled.opacity || theme.button.default.disabled.opacity};
+            padding: ${({theme, type}) => theme.button[type].disabled.padding || theme.button.default.disabled.padding};
+        }
+    }
+`;
```
codeviking: Woah! That's a lot!

Are we certain we need to output all of this?  Is this another spot where we're unnecessarily outputting styles that already exist in the css?

ui/src/lib/theme/button.ts
```diff
@@ -0,0 +1,141 @@
...
+    contrastColor?: Color;
+    border?: string;
+    contrastBorder?: string;
+    background?: Color;
+    textTransform?: string;
+    focus: any; // force required so we dont have to check for undefined later
+    hover: any; // force required so we dont have to check for undefined later
+    disabled: any; // force required so we dont have to check for undefined later
+}
+interface Button {
```
codeviking: Maybe `ButtonType` for clarity, as `Button` is a bit overloaded?

ui/src/lib/theme/button.ts
```diff
@@ -0,0 +1,141 @@
...
+    primary: ButtonStates;
+    default: ButtonStates;
+    link: ButtonStates;
+    marketing: ButtonStates;
+}
+export const defaultSpacing = spacing.md;
+const paddingWithBorderInside = (borderWidth: number, initilaPad = defaultSpacing) => {
+    return `0 ${convertPixelsToRem(parseFloat(convertRemStrToPx(initilaPad.rem))-borderWidth)}`;
+}
```
codeviking: Woah, this seems a little complex.

We should have a rule in our global CSS that sets `box-sizing` to `border-box`:

```
* {
   box-sizing: border-box;
}
```

That should obviate the need for this type of thing.  Also, `initilaPad` is likely supposed to be `initialPad` or `initialPadding` even, for clarity?

ui/src/lib/components/button.tsx
```diff
@@ -0,0 +1,75 @@
...
+            }};
+            color: ${({theme, type, contrast}) => {
+                return !contrast
+                    ? theme.button[type].disabled.color || theme.button.default.disabled.color
+                    : theme.button[type].disabled.contrastColor || theme.button.default.disabled.contrastColor || theme.button[type].disabled.color || theme.button.default.disabled.color
+            }};
+            opacity: ${({theme, type}) => theme.button[type].disabled.opacity || theme.button.default.disabled.opacity};
+            padding: ${({theme, type}) => theme.button[type].disabled.padding || theme.button.default.disabled.padding};
+        }
+    }
```
codeviking: I think we should try to reduce the amount of boiler plate here, as this is going to get pretty onerous to maintain as time goes.  In the theme definition below we more or less have the same code that we're putting here -- so it seems like this code could be closer to something like:

```
const Button = styled(AntButton)`
   // Common stuff
   outline: none;

   ({ variant }) => withStylesForVariant(variant)
`;

The idea being `withStylesForVariant` would pull styles from a definition somewhere that provides a way to define styles per variation.  This would avoid all the inline comparison / theme references that make this tricky to read.

This is a tad closer to the library you referenced.  That said I also don't think it'd be a crazy amount of code to write, so it might be worth coming up with a home grown solution. 

Also, related to this -- I'm not sure if we want to let people adjust the specifics of each variant (i.e. we may not need to expose `Button.marketing.fontWeight` as something they can override.  The variations we provide in my mind could be prepackaged and non-overridable.  I like this because it keeps the theme API simple.  The more complex it gets the less likely *our* userbase is to use it (as they're likely a bit less eager to embrace complexity in the interest of flexibility).

Let's chat about this in person tomorrow -- I feel like now is the right time to experiment a bit here as to nail down the right API that makes this stuff easy to maintain while providing the desired level of flexibility.

File: ui/src/lib/theme/spacing.ts [hunk 7]
```diff
@@ -1,10 +1,20 @@
 import { convertPixelsToRem } from '../utils/base';
 
+export class Spacing {
+    public rem: string;
+    constructor(public displayName: string, public px: number) {
+        this.rem = convertPixelsToRem(px);
+    }
+    toString() {
+        return this.rem;
+    }
```
codeviking: This should probably include the unit, for clarity:

```
return `${this.rem}rem`
```

ui/src/lib/theme/spacing.ts
```diff
@@ -1,10 +1,20 @@
 import { convertPixelsToRem } from '../utils/base';
 
+export class Spacing {
+    public rem: string;
+    constructor(public displayName: string, public px: number) {
+        this.rem = convertPixelsToRem(px);
+    }
```
codeviking: One thought.  With data models like this I usually expose all fields in the constructor, and then create factory methods for deriving fields. This allows additional factories to be added later without changing the class implementation.  For instance:

```
class Spacing {
   constructor(public displayName, public px: number, public rem: number) {}
   static fromPixels(displayName: string, px: number) {
      return new Spacing(displayName, px, convertPixelsToRem(px))
    }
}
```

ui/src/demo/pages/Design/Buttons.tsx
```diff
@@ -2,13 +2,156 @@ import * as React from 'react';
...
     render() {
         return (
-            <Paragraph>Buttons coming soon...</Paragraph>
+            <React.Fragment>
+                <PageTitle>Buttons</PageTitle>
+
+                <h3>Appearance and Behavor</h3>
+                <BodySmall>
+                    Basic buttons come in four style types: primary, default, link, and marketing.
+                </BodySmall>
```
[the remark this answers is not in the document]
jonborchardt: im lost
bodysmall is a font and a size (https://varnish.staging.apps.allenai.org/design/typography)

ui/src/demo/pages/Design/Buttons.tsx
```diff
@@ -2,13 +2,156 @@ import * as React from 'react';
...
+
+                <SectionDivider />
```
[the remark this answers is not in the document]
jonborchardt: great idea!

ui/src/demo/pages/Design/Buttons.tsx
```diff
@@ -2,13 +2,156 @@ import * as React from 'react';
...
+        )
+    }
+}
+
+const ButtonGrid = styled.div`
+    display: grid;
+    grid-template-columns: repeat(2, max-content);
+    align-items: center;
+    grid-gap: ${({theme}) => theme.spacing.lg};
+`;
```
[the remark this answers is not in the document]
jonborchardt: https://github.com/allenai/varnish/issues/36

ui/src/lib/components/button.tsx
jonborchardt: this is all needed IFF we want to be able to see the values in the theme
all of these properly override the ant style
(also we could remove some if nick says the ant default is good enough)

ui/src/lib/theme/button.ts
jonborchardt: yup

ui/src/lib/theme/button.ts
jonborchardt: we are already setting box-sizing (although, ill add it at the root for clarity)
the padding change here is to accommodate 3 levels of border... ill make it cleaner

ui/src/lib/theme/spacing.ts
jonborchardt: great idea

ui/src/lib/theme/spacing.ts
jonborchardt: rem includes unit

PR #42, opened 2019-06-26
Jon/header

jonborchardt:


ui/src/demo/Home.tsx
```diff
@@ -35,19 +37,20 @@ export default class Home extends React.PureComponent<RouteComponentProps> {
...
-        }
+        }*/
     ];
 
     render() {
         return (
             <TransparentLayout>
                 <Header>
-                    <AI2Logo />
+                    <HeaderLogo logo={logoWithText} wideWidth={"194px"} skinnyWidth={"72px"} height={"56px"} alt="Varnish" />
```
codeviking: Interesting. What are these widths for? I'd expect them to just be controlled by breakpoints?

Also, if we're setting `height`, we shouldn't be. We should use `position: sticky`, etc.

ui/src/lib/components/AI2Banner.tsx
```diff
@@ -0,0 +1,38 @@
...
+    max-width: ${({theme}) => theme.breakpoints.xl};
+    margin: 0 auto;
+    padding: ${({theme}) => `0 ${theme.spacing.lg}`};
+`;
+
+const Logo = styled.img`
+    display: block;
+    max-width: 100%;
+    height: auto;
+`;
```
codeviking: Hmm, why are we setting `height` / `width` here and above? I think this overrides the `width` and `height` we set above.

Ideally we'd use the native, `width` and `height` provided by the browser and avoid setting dimensions in CSS when possible.

ui/src/lib/components/Header.tsx
```diff
@@ -0,0 +1,100 @@
...
+        )
+    }
+}
+
+const Content = styled.div`
+    && {
+        display: flex;
+        align-items: center;
+        margin: auto;
+        height: 80px;
```
codeviking: Why are we setting `height`?

ui/src/lib/components/Header.tsx
```diff
@@ -0,0 +1,100 @@
...
+
+const Sticky = styled(Layout.Header)`
+    && {
+        position: sticky;
+        top: 0;
+        background: white;
+        z-index: 1;
+        width: 100%;
+        box-shadow: 0px 4px 16px rgba(10, 41, 57, 0.08);
+        height: auto;
```
codeviking: Do we need this? Won't it just be sized by it's content?

ui/src/lib/components/HeaderLogo.tsx
```diff
@@ -0,0 +1,45 @@
...
+    return (
+        props.alwaysShowText
+            ? fullLogo
+            : <ResponsiveLogoContainer
+                breakWidth={props.breakWidth}
+                width={props.skinnyWidth}
+                height={props.height}
+            >{fullLogo}</ResponsiveLogoContainer>
+    );
+};
```
codeviking: I think the name is a bit too general for my taste.  What this really is is a "SlidingWindow" that'll crop an image at a given breakpoint to provided dimensions.

So, rather than `HeaderLogo` maybe we call this `SlidingWindow` or something? Naming is hard -- I suppose my point is I could see this being used in another scenarios outside of the header.

Maybe it's over complicating things by giving it a magic name like that -- feel free to pick a name you think is right (if one comes to mind) or leave as is.

ui/src/lib/components/HeaderLogo.tsx
```diff
@@ -0,0 +1,45 @@
...
+ */
+export const HeaderLogo = (props: Props) => {
+    const fullLogo = (
+        <img src={props.logo}
+                width={props.wideWidth}
+                height={props.height}
+                alt={props.alt} />
+    );
+    return (
+        props.alwaysShowText
```
codeviking: I'm not sure I follow the intent of this parameter.

If I always wanted to show the text, wouldn't I just use a `` tag, and avoid using `HeaderLogo` all together?

ui/src/lib/components/HeaderLogo.tsx
```diff
@@ -0,0 +1,45 @@
...
+            : <ResponsiveLogoContainer
+                breakWidth={props.breakWidth}
+                width={props.skinnyWidth}
+                height={props.height}
+            >{fullLogo}</ResponsiveLogoContainer>
+    );
+};
+
+const ResponsiveLogoContainer = styled.div<{width: string, height: string, breakWidth?: string}>`
+    display: flex;
```
codeviking: Why is `flex` necessary?

ui/src/lib/components/HeaderLogo.tsx
```diff
@@ -0,0 +1,45 @@
+import * as React from 'react';
+import styled from 'styled-components';
+
+interface Props {
+    logo: string;
```
codeviking: I'd just call this `src` since that's what it is.

The word `logo` leaves me wondering -- could this be HTML?

ui/src/demo/Home.tsx
jonborchardt: these widths are the size of the image.... and the size of the sliding window holding the image
they cannot be breakpoints

i was only setting height, because you advised me to... i like to just set width and auto height, but you told me to set both on images... please advise

ui/src/lib/components/Header.tsx
jonborchardt: because we want to ensure that the header is a consistent height even if the user does not add any links and does not add a logo

in other words, if we just rely on padding, then if the user has a smaller logo and does not add any links... like grover (no logo, just text, no links)
then the header would shrink....

please advise

ui/src/lib/components/Header.tsx
jonborchardt: unfortunately, the height of the ant.header shrinks smaller than the content unless i set this to auto
image

ui/src/lib/components/HeaderLogo.tsx
jonborchardt: deal

ui/src/lib/components/HeaderLogo.tsx
jonborchardt: this is copied from your code....
it allows user to force the window open... ill remove if you dont want it

ui/src/lib/components/HeaderLogo.tsx
jonborchardt: removed... was vestigial

ui/src/lib/components/HeaderLogo.tsx
jonborchardt: yup, good

ui/src/demo/Home.tsx
codeviking: When using an `` tag, best practice is to set the width and height via the attributes, and not alter the dimensions with CSS -- so in this case we'd in theory need to be able to adjust the width and height using JavaScript that alters the rendered output when a breakpoint is triggered. Do we have that capability?  Something like:

```
<MediaQuery query={`media screen and (${Breakpoint.TABLET))`}>({ isActive }) => (
   <img width={isActive ? 100 : 200} height={isActive ? 300 : 100 } />
</MediaQuery>
```

If we can't do that, then as is is the right way to go -- it's just a bit confusing as we set dimensions both in the HTML and CSS.  There's no huge downside other than overhead for the developer to be aware of / process.

ui/src/lib/components/AI2Banner.tsx
codeviking: Per above, it might be necessary.  Feel free to do what works; Clearly I'm having a hard time recalling my own rational for the code.

Maybe a comment explaining things might be helpful. Something like:

```
// We set the dimensions as HTML attributes so that it renders correctly 
// without CSS being applied, but use CSS to ensure the image changes
// dimensions at different breakpoints
```

ui/src/lib/components/Header.tsx
codeviking: Ah, so it's about design fidelity (i.e. we want the header to always be `xx`px).  That said, this means that if the content wraps -- i.e. grows larger than the header on a small resolution -- the header won't grow encompass the content. This typically makes things look really weird. It also makes the layout less accessible to the changes people make (i.e. if a person wants 3 lines of text in the header, should the header grow to capture that content, or should the text just overflow and look crappy)?

I tend to believe that UI components should always be driven (and sized by) their content, as I think it makes them more flexible for the end users.

That said, this is ultimately a question of the design intent.  If we intend to mandate that a header should always be a provided height, regardless of what's in it, and we want to discourage people who try to operate outside of these boundaries, then as-is is the right call -- I just don't agree with that concept, nor think it leads to a flexible, well written component library.  That said, this isn't my call -- it's up to you and Nick -- so go ahead and to what y'all think is right!

ui/src/demo/Home.tsx
jonborchardt: we dont set the img attrs AND css
we set the img attrs on the image
and css on the div around the image....

this is you code :-P

my q, is... should we set height and with, or just one and let the other auto?

ui/src/lib/components/AI2Banner.tsx
jonborchardt: i removed it in this case

ui/src/lib/components/Header.tsx
jonborchardt: ill set it to min-height... sound good?

ui/src/lib/components/Header.tsx
codeviking: Interesting.  That shouldn't be happening, and isn't necessary on the Incubator. I'd encourage you to look into this more, and/or we can take a look tomorrow.

I think it's worth resolving, but feel free to overrule. Usually these types of oddities are indicative of some sort of layout issue.

jonborchardt: incubator does not use ANT.HEADER
this is an issue with ant

ui/src/lib/components/HeaderLogo.tsx
codeviking: Heh, awesome. 

I feel like we should lift it up, as noted -- a user could do something like:

```
alwaysShowText ? <img /> : <ResponsiveWindow .. ./>
```

Feel free to make a call though.

ui/src/lib/components/HeaderLogo.tsx
codeviking: Ah, the image needs `display: block;`, I think.  Try that.

ui/src/lib/components/HeaderLogo.tsx
jonborchardt: i remove it... user can do it outside of varnish as you show

ui/src/lib/components/Header.tsx
codeviking: Sure, though as noted, I think this is ultimately a call that's up to you and Nick. We should clarify the design goal.

In other words -- if a user wants their header to be smaller (because they have less content, or a small logo), do we want the header to still be large and the content to be vertically centered, etc)?

In other words, I'd walk through the variations with Nick and implement the behavior he thinks is right. I'd try to take into consideration the variations in layout users might want -- so that the controls are as flexible as they can be without requiring our help / input (as the more they can do without us, the better).

ui/src/lib/components/HeaderLogo.tsx
jonborchardt: fixed, ty

ui/src/lib/components/Header.tsx
codeviking: Ah, bummer.  Ok, I'd be interested in digging into this, but if it's not worth it that's ok too.  Does `ant` set a height that we're overriding, maybe? If that's the case a comment here would be worth it:

```
height: auto; /* Ant sets a height, we need to unset that. */
```

Also, I think `height: initial` might be correct and have the same effect -- it sets the style to that which is applied by default, effectively "undoing" ant's setting.

ui/src/demo/Home.tsx
codeviking: Heh, man.  I'm reviewing my own code :).

What are the css settings doing? Are the dimensions of the logo changing at various breakpoints? If not maybe we don't need it at all, and can just set the `height` and `width` of the image.

ui/src/lib/components/AI2Banner.tsx
codeviking: If it works, great!

ui/src/lib/components/Header.tsx
jonborchardt: you are convincing me....

ui/src/demo/Home.tsx
jonborchardt: we dont want to squish the image...
i think this code is correct now...
```
export const ResponsiveWindowImage = (props: Props) => {
    const fullImage = (
        <img src={props.src}
            width={props.wideWidth}
            height={props.height}
            alt={props.alt} />
    );
    return (
        <ResponsiveImageContainer
            breakWidth={props.breakWidth}
            width={props.skinnyWidth}
            height={props.height}>
            {fullImage}
        </ResponsiveImageContainer>
    );
};

const ResponsiveImageContainer = styled.div<{width: string, height: string, breakWidth?: string}>`
    img {
        display: block;
    }
    @media (max-width: ${({theme, breakWidth}) => breakWidth ? breakWidth : theme.breakpoints.sm}) {
        width: ${({width}) => width};
        height: ${({height}) => height};
        overflow: hidden;
    }
`;
```

codeviking: 👍

ui/src/lib/components/Header.tsx
jonborchardt: initial works!
and added a comment

PR #101, opened 2019-07-30
Add basic table

meslater1030:
This PR:

* Adds the simplest possible table from Ant Design to Varnish
* Provides a demo page that looks like the following:

Screen Shot 2019-07-30 at 2 42 41 PM

### DIFF
--- a/ui/src/lib/components/Table.tsx
+++ b/ui/src/lib/components/Table.tsx
@@ -2,7 +2,7 @@ import styled from 'styled-components';
 import { Table as AntTable } from 'antd';
 
 export const Table = styled(AntTable)`
-  background-color: ${props => props.theme.color.N1.hex};
-  border: 1px solid ${props => props.theme.color.N4.hex};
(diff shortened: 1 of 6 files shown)


ui/src/lib/components/Table.tsx
```diff
@@ -0,0 +1,8 @@
+import styled from 'styled-components';
+import { Table as AntTable } from 'antd';
+
+export const Table = styled(AntTable)`
+  background-color: ${props => props.theme.color.N1.hex};
```
jonborchardt: you dont need to add .hex

jonborchardt: should be palette.background.light

ui/src/lib/components/Table.tsx
```diff
@@ -0,0 +1,8 @@
...
+  border: 1px solid ${props => props.theme.color.N4.hex};
```
jonborchardt: should be palette.border.main

PR #109, opened 2019-08-09
Ensure the footer and page share the same background color.

codeviking:
This modifies the `<Footer />` so that it sets the background
of the page. This fixes the issue where the footer appears
to "hang" on the page on larger screens without a lot of content.

I added an option that lets users disable this behavior, in
the event that it causes artifacts / issues or they'd like to
set the page background separately.

Before:

image

After:

image

We should make sure @Pelagic-sky wants this behavior, but I *think* it's what's desired.


ui/src/lib/components/shared.tsx
```diff
@@ -47,8 +51,8 @@ export const LeftSider = styled(Layout.Sider).attrs({
 export const PaddedContent = styled(Layout.Content)`
     && {
         max-width: ${({theme}) => theme.breakpoints.xl};
-        margin: 0 auto 4rem auto;
-        padding: ${({theme}) => `0 ${theme.spacing.lg}`};
+        margin: 0 auto;
+        padding: ${({theme}) => `0 ${theme.spacing.lg} ${theme.spacing.xxl}`};
```
codeviking: This felt simpler, and more correct (I usually use padding for this purpose).  That said if this isn't desirable for some reason let me know and I can reverse it.

ui/src/lib/components/Footer.tsx
```diff
@@ -9,13 +10,23 @@ export type FooterVariant = 'default' | 'dark';
...
+        setPageBackground: true
+    }
     render() {
         let contrast = this.props.variant === 'dark';
         return (
             <StyledFooter contrast={contrast}>
+                {this.props.setPageBackground ? (
+                    <WithPageBackground
+                        background={contrast
+                            ? palette.background.dark.hex
```
jonborchardt: should not need .hex

codeviking: Ah, I just have to change the type of the property to `Color` instead of `string`. 👍

Issue #103, opened 2019-08-05
I have to import components in isolation if I use a different router.


codeviking:
In a NextJS application one can't import components like so:

```
import { Button, Header } from "@allenai/varnish/components"
```

This fails as the `react-router-dom` dependency isn't installed (which the various link related components depend upon).

I can workaround this for now by importing them individually:

```
import { Header } from "@allenai/varnish/components/Header";
import { Button } from "@allenai/varnish/components/Button";
```

That said this is a bit of a pain.

jonborchardt:
im not sure there is work to be done here with the fork

PR #110, opened 2019-08-09
Fix the footer so it works with non react router projects.

codeviking:
This fixes the `<Footer />` component by ensuring the import
path for `<ExternalLink />` doesn't rely on `react-router`.
This enables the `<Footer />` to be used in projects that
don't include `react-router`.

The components still have to be imported in isolation (i.e. without
using the barrel file, re #103). We may accordingly come up
with a more pluggable mechanism and/or split `react-router` tied
functionality into a separate module in the future.


ui/src/lib/components/link/InternalLink.tsx
```diff
@@ -0,0 +1,6 @@
+import styled from 'styled-components';
+import { NavLink } from 'react-router-dom';
+
+import { baseLinkStyles } from '.'
```
jonborchardt: i think you meant './base'

codeviking: Good catch, fixed!

PR #120, opened 2019-08-19
pr fixes and made left menu move up on scroll

jonborchardt:


ui/src/demo/Home.tsx
```diff
@@ -36,35 +37,33 @@ export default class Home extends React.PureComponent<RouteComponentProps> {
...
-                    <Spacer />
-                    <TopMenu
-                        defaultSelectedKeys={[this.props.location.pathname]}>
-                        {this.routes.map(({ path, label }) => (
-                            <TopMenuItem key={path}>
-                                <InternalLink to={path}>{label}</InternalLink>
-                            </TopMenuItem>
-                        ))}
-                    </TopMenu>
+            <BackgroundLayout color="white">
```
codeviking: 🤘 This is great.

ui/src/demo/pages/Shared/LeftMenuPage.tsx
```diff
@@ -59,29 +59,29 @@ export class LeftMenuPage extends React.PureComponent<RouteComponentProps & Prop
...
+                                                        {label}
+                                                    </BodySmall>
+                                                </Wrapping>
+                                            </IconMenuItemColumns>
+                                        )
+                                        : (
+                                            <React.Fragment>
+                                                <Icon type={icon} /><span>{label}</span>
+                                            </React.Fragment>
+                                        )
```
codeviking: We should get `prettier` in sooner rather than later.  Should we cut a ticket to add prettier, and add CI that ensures PRs don't have formatting issues?

ui/src/lib/components/Columns.tsx
```diff
@@ -1,48 +1,20 @@
...
 
-export const Columns = styled.ul<{count: number}>`
+export const Columns = styled.ul<{count?: number, gridTemplateColumns?: string, breakpoint?: string}>`
     display: grid;
-    grid-template-columns: repeat(${({count}) => count}, 1fr);
+    grid-template-columns: ${({gridTemplateColumns, count}) => gridTemplateColumns
+        ? gridTemplateColumns
+        : (count
+            ? `repeat(${count}, 1fr)`
+            : undefined)};
```
codeviking: Nice.

ui/src/lib/components/Layout.tsx
```diff
@@ -2,64 +2,120 @@
...
-            padding: ${({theme}) => `0 ${theme.spacing.sm}`};
+    onScroll = () => {
+        if (this.elRef.current !== null) {
+            const distance = window.scrollY - this.lastScrollY;
+            this.lastScrollY = window.scrollY;
+            if ( distance < 0 ) {
+                this.setState({ isCollapsed: false });
+            } else {
+                this.setState({ isCollapsed: true });
+            }
```
codeviking: Can the user click to expand it, even if they're scrolled down the page? As is it looks like they can only expand it by scrolling up? Is that desired?

Also, you could write this as:

```
const isCollapsed = distance < 0;
this.setState({ isCollapsed })
```

ui/src/lib/components/Layout.tsx
```diff
@@ -2,64 +2,120 @@
...
+    getStyle() {
+        let top = 113; // magic numbers are from inspecting the header.... we could pass them in
+        if (!this.props.alwaysVisible && this.state.isCollapsed && this.elRef.current !== null) {
+           top = 84;
+        }
+        return {
+            top: convertPixelsToRem(top),
+            height: `calc(100vh - ${convertPixelsToRem(top)})`,
+            transition: 'top 200ms ease-in-out, height 200ms ease-in-out'
+         };
```
codeviking: So, `transition` is never changing -- should it really be here?

Also, I'm pretty confused as to why we need this.  Let's follow up in person tomorrow.  I'm inclined to think we can do this all with `css`, provided we use `position: sticky` instead of `fixed`, but I could be wrong.

ui/src/demo/pages/Shared/LeftMenuPage.tsx
jonborchardt: i think we have one, ill cut if we dont

ui/src/lib/components/Layout.tsx
jonborchardt: neither this not the header expand on click...

ui/src/lib/components/Layout.tsx
jonborchardt: im supper happy to have help figuring it out!!... but it is needed afaik

codeviking: Sure.  Let's look at it tomorrow if there's time.

At the very least we'll want to somehow calculate this by sharing the `Header` `ref` with this component, or something...or maybe some sort of hidden `Context` we use to pass layout dimensions to different components that need it.

Anyway, we definitely don't want it to be hard-coded, furthermore isn't our header responsive to content height (i.e. if someone puts a big logo in, won't it grow to accomodate it)?

PR #123, opened 2019-08-20
Jon/type ant table

jonborchardt:


ui/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@allenai/varnish",
-  "version": "0.3.22",
+  "version": "0.4.0",
```
codeviking: Shouldn't you wait to increment this until the actual release is drafted, so that the version change + tag point to the SHA derived from master?

ui/src/lib/components/Table.tsx
```diff
@@ -1,8 +1,24 @@
...
+export interface TableProps<T> extends AntTableProps<T>{};
+export interface TableState<T> extends AntTableState<T>{};
+
+export class Table<T> extends React.PureComponent<TableProps<T>, TableState<T>> {
+    render() {
+        return (
+            <StyledTable {...this.props} />
+        )
+    }
+}
```
codeviking: I might add some docs explaining why this trickery is necessary (we'll likely forget down the road).

ui/package.json
jonborchardt: 0.4.0 is already released... but i forgot to check it in
after this goes in, i will be moving to 0.4.1 with no code change...

PR #136, opened 2019-08-29
Fix #129; Add h6, and shuffle header font sizes.

### LINKED ISSUE #129: Shuffle H specs to include a new H3 size
Currently we do not take advantage of the H6 size. So we're introducing a new size to fill out the spec. 

All current sizes from H3 to H5 will shift down to make room for a new H3. So H5 becomes H6 and H4 becomes H5 and H3 becomes H4. 

New H3 specs:
Raleway Bold, 30px, Line Height 36px.

codeviking:
Here's how things look:

image


jonborchardt:
reviewing code, but from the image, it looks wrong.... look at h5 and h6

codeviking:
Good catch, thanks.

codeviking:
@jonborchardt Updated, take another look.

Issue #98, opened 2019-07-29
Add Tabs to Varnish demo


Pelagic-sky:
I took a look at the stock Ant tabs and I like their design. However, I do not like the animation of the content sliding left and right on click. I would like to kill the animation and then see how tabs are implemented to make sure their font size doesn't need to be adjusted.

codeviking:
@Pelagic-sky Is there an icon you'd like to use for the left-hand navigation for this page?

![image](https://user-images.githubusercontent.com/801451/63954946-cbf01900-ca38-11e9-8d46-a054daf5586c.png)

jonborchardt:
all the icons in the demo are tempo, so just pick one: https://ant.design/components/icon/
@Pelagic-sky has a ticket to make new icons for demo

codeviking:
Ah! Cool. Will do.  Thanks!

PR #149, opened 2019-08-29
Re #98, #147: Add tabs to the demo.

codeviking:
This adds the `antd.Tabs` component to the demo with a few
small modifications:

* We disable the animation by default
* We set the font-size to be inherited, rather than a fixed
  `14px` (which is odd).

This wraps the `antd` component in a different fashion, as
to prevent the loss of typing information so that it can be
easily used in the same fashion as the default component produced
by `antd`.

I'm not going to merge until we've all signed off on whether we
like this.

I also still need to run over the UI with @Pelagic-sky to assess
whether there's further styling adjustments to make.


codeviking:
Here's a preview environment. @Pelagic-sky feel free to take a look, you can either request changes here asynchronously or let me know what you'd like changed tomorrow.

https://varnish.beafc9.apps.allenai.org/design/tabs

ui/src/demo/pages/Design/Design.tsx
```diff
@@ -14,7 +14,8 @@ import {
     Palette,
     Spacing,
     Typography,
-    Tables
+    Tables,
+    TabsPage
```
jonborchardt: Should we convert them all to be xPage?

codeviking: Good question. I only renamed it because I'm importing `Tabs`, so there was a conflict.

I could:

* Just not import the `Tabs` class. I noticed some pages include examples, while others don't (they just use the live examples). What's our desired state moving forward?  Doing this would allow me to rename `Design/TabsPage` back to `Design/Tabs`.p
* Rename all pages to include `Page`, but that seems like it's not worth the effort.
* Leave it as is, which isn't great as there's inconsistency in the nomenclature.

What do you think?

jonborchardt: ill add a ticket to add Page to all... its simple and has no downside

codeviking: Sure, but the question remains -- do we want examples on the page, or not?  I could just use the live example, as some pages do, and then this problem goes away (I can rename it to `Tabs`).

Either way is fine by me. LMK what you think.

Pelagic-sky:
@codeviking Per our convo, the font size coming through is 14px when it should be 16px.

ui/src/demo/pages/Design/TabsPage.tsx
```diff
@@ -0,0 +1,53 @@
...
+        <SectionWithDivider>
+            <h3>Usage</h3>
+            <DefaultLiveProvider code={basicUsage} />
+        </SectionWithDivider>
+    </React.Fragment>
+);
+
+const basicUsage = `
+/* Uncomment for use:
+import { Tabs } from '@allenai/varnish/components';
```
meslater1030: Thanks for this!

ui/src/demo/pages/Design/Design.tsx
codeviking: I reverted this, we can always reimplement in the future if desired!

ui/src/demo/pages/Design/TabsPage.tsx
codeviking: 🤘

Issue #96, opened 2019-07-29
Update dropdown icon


Pelagic-sky:
I've designed a dropdown icon to replace the default ant one uses.
You can find it here: https://drive.google.com/drive/folders/1UUeVIs4Jh1ddd0A7sxJxeyv0HGOLYrf1?usp=sharing

It's file name is icon-form-dropdown.svg

PR #152, opened 2019-08-30
Fix #95, #96; Update icons for `<Select />` and `<Datepicker />`.

### LINKED ISSUE #95: Update calendar icon
I've designed a calendar icon to replace the default ant one used in inputs.
You can find it here: https://drive.google.com/drive/folders/1UUeVIs4Jh1ddd0A7sxJxeyv0HGOLYrf1?usp=sharing

It's file name is icon-form-calendar.svg

codeviking:


ui/src/lib/components/form-calendar.svg
```diff
@@ -0,0 +1 @@
+<svg fill="none" height="14" viewBox="0 0 14 14" width="14" xmlns="http://www.w3.org/2000/svg"><g fill="#8d939c"><path clip-rule="evenodd" d="m13 2h-12v11h12zm-1 1h-10v9h10z" fill-rule="evenodd"/><path d="m1 5h12v1h-12z"/><path d="m4 1h1v3h-1z"/><path d="m9 1h1v3h-1z"/></g></svg>
```
jonborchardt: i propose we dump these into an components/icon folder
instead of having them scattered in components...

or maybe even lib/icons with the thought that we could eventually serve them?

codeviking: Sure. 👍

Issue #70, opened 2019-07-15
add z-index list to theme


jonborchardt:

codeviking:
Complete. We have #153 open to give them better names.

PR #154, opened 2019-08-30
Fix #90, #70; Z-index related fixes.

### LINKED ISSUE #90: The date picker icons are at a higher z-index than our header.
The icon to the right of the logo is supposed to be underneath it:

image

codeviking:
This adds `zIndex` to our theme, and ensures they're passed through
to ant.

We also set a `zIndex` for our header which is above dropdowns
and datepickers, as to ensure they don't show up over the header.


ui/public/theme.css
```diff
@@ -1318,28 +1318,28 @@ html {
   line-height: 22px;
 }
 .ant-alert-success {
-  background-color: #C1F7E6;
+  background-color: #E4FFF7;
```
codeviking: @jonborchardt I'm not sure why these changed, let me know what I messed up :).

ui/src/lib/theme/zIndex.js
```diff
@@ -0,0 +1,27 @@
...
+    modal: 1000,
+    modalMask: 1000,
+    message: 1010,
+    notification: 1010,
+    popover: 1030,
+    dropdown: 1050,
+    picker: 1050,
+    tooltip: 1060,
+    header: 1070
+};
```
codeviking: Any idea while this file got generated? Does this happen when I run `yarn build`?

PR #158, opened 2019-08-30
Refactor layout mechanism.

codeviking:
~~This still relies on ant's Layout, but reworks things a
bit in the demo as to help clarify and move towards a canonical
API.~~

**This is now a large-scale refactor of the layout mechanism, as to 
    replace ant's mechanism with our own that's grid based. This solves
    several UI bugs and (ideally) simplifies the API for end users.**

This adds a `Layout` context, so that we only have to configure
the `LayoutVariant` in a single spot. Components that need it
can now use `Layout.Consumer` to retrieve the current `LayoutVariant`.

As part of this, I also made some high-level changes to the structure
and nomenclature in the demo in the interest of clarity.  Apologies
for how large this change ended up being.

Lastly, I cleaned up a few stray issues I noticed along the way:

* Deep navigation to a design subpage now properly activates the design
   navigation menu item in the header.
* Fixed the gap between the header and the left hand navigation menu.

Given the extent of the changes I figured I'd deploy an adhoc environment
for you to verify.  Please take a look and let me know if I broke anything: 

Here's an adhoc environment for testing / verification purposes: 
https://varnish.layout.apps.allenai.org


ui/src/demo/components/shared/PageWithLeftMenu.tsx
```diff
@@ -0,0 +1,49 @@
...
+    children: [ WithCollapsible, React.ReactNode ]
+}
+
+interface State {
+    menuCollapsed: boolean;
+}
+
+export class PageWithLeftMenu extends React.PureComponent<Props, State> {
+    siderWidthExpanded = '12rem';
+    siderWidthCollapsed = '5rem';
```
codeviking: I'd like to derive this using refs, if possible.  Given the `LeftSider` is a child of this component, I should be able to do this easily.

ui/src/demo/components/Header.tsx
```diff
@@ -0,0 +1,57 @@
...
+    TopMenuItem,
+    HeaderColumns
+} from '../../lib/components';
+import packageJson from '../../../package.json';
+import logoWithText from './varnishLogo.svg';
+import { AppRoute } from '../AppRoute';
+import { Welcome, Design } from '../pages';
+
+const routes: AppRoute[] = [
+    {
```
jonborchardt: why does this exist here ant in app ln 11?

ui/src/demo/components/Header.tsx
```diff
@@ -0,0 +1,57 @@
...
+];
+
+interface Header {
+    location: Location;
+}
+
+export const Header = ({ location }) => (
+    <VarnishHeader>
+        <HeaderColumns gridTemplateColumns="auto auto 1fr auto">
+            <ResponsiveWindowImage
```
jonborchardt: this is going away as a concept... we should just have a logo and text, not a logo that includes text

ui/src/demo/components/shared/PageWithLeftMenu.tsx
```diff
@@ -0,0 +1,49 @@
...
+        super(props);
+        this.state = { menuCollapsed: false }
+    }
+    handleMenuCollapse = () => {
+        this.setState({ menuCollapsed: !this.state.menuCollapsed });
+    }
+    render() {
+        const [ renderMenu, main ] = this.props.children;
+        return (
+            <Layout variant="app">
```
jonborchardt: i thought this was supposed to be coming through the context now?

ui/src/demo/pages/Welcome.tsx
```diff
@@ -0,0 +1,54 @@
+import * as React from 'react';
+import { RouteComponentProps } from 'react-router';
+
+import { Header } from '../components';
+import { ExternalLink, Body, Layout, MaxWidthCenteredContent, Page, Footer } from '../../lib/components';
+
+export const Welcome = (props: RouteComponentProps) => (
+    <Layout variant="app" bgcolor="white">
```
jonborchardt: shouldn't this be coming via the contest now?

ui/src/demo/pages/Welcome.tsx
```diff
@@ -0,0 +1,54 @@
...
+        <Header location={props.location} />
+        <Layout variant="default">
```
jonborchardt: outer variant is app and inner is default... this is confusing

ui/src/demo/pages/design/Design.tsx
```diff
@@ -0,0 +1,178 @@
...
+    {
+        path: `${parentPath}/tabs`,
+        label: 'Tabs',
+        component: Tabs,
+        iconSrc: tabIcon
+    }
+];
+
+export const Design = (props: RouteComponentProps) => (
+    <Layout variant="app" bgcolor="white">
```
jonborchardt: maybe our varients should just be
'leftAlign' | 'centerAlign' ?

jonborchardt: or just
Layout: {align: 'left' | 'center', ...}

ui/src/demo/pages/design/Design.tsx
```diff
@@ -0,0 +1,178 @@
...
+        <Header location={props.location} />
+        <PageWithLeftMenu>
+            {isCollapsed => (
```
jonborchardt: isMenuCollapsed

ui/src/lib/components/Footer.tsx
```diff
@@ -33,29 +34,32 @@ export class Footer extends React.PureComponent<Props> {
     render() {
         const contrast = this.props.variant === 'dark' ? true : undefined
         return (
-            <StyledFooter
+            <Layout.Consumer>{layout => (
```
jonborchardt: i wouyld newline this for clarity

ui/src/lib/components/Layout.tsx
```diff
@@ -90,26 +96,13 @@ export class LeftSider extends React.PureComponent<LeftSiderProps, LeftSiderStat
         window.removeEventListener("scroll", this.onScroll);
     }
 
-    getStyle() {
-        let top = 125; // magic numbers are from inspecting the header.... we could pass them in
-        if (!this.props.alwaysVisible && this.state.isCollapsed && this.elRef.current !== null) {
```
jonborchardt: where do you do this now?
this fixed a bad looking left menu issue

ui/src/demo/components/Header.tsx
codeviking: Good question. Clearly they should be consolidated. I likely unwound this and forgot to re-merge them. Good catch.

I'll consolidate to a single definition.

ui/src/demo/components/Header.tsx
codeviking: Cool, good to know, but I think I'll leave as is for now. Seems like a separate effort.

ui/src/demo/components/shared/PageWithLeftMenu.tsx
codeviking: Well, the `Layout` element "provides" the context.

I figured end users would use this to declare what layout is "active" at that point in the tree. Both creating the required DOM and ensuring anything that's descent of `Layout` (without another, closer`Layout` parent in it's ancestors) can "consume" the layout set by that component to determine what layout it's being rendered into.

Let's chat about this tomorrow in person as it's a bit nuanced and accordingly deserves some thought.

ui/src/demo/pages/Scratch.tsx
```diff
@@ -0,0 +1,92 @@
...
+                    <Footer />
+                </Content>
+            )}
+        </Layout>
+    );
+}
+
+const Title = styled(HeaderTitle)`
+    padding: ${({ theme }) => theme.spacing.lg} 0;
+`;
```
codeviking: @jonborchardt I had already taken a stab at replacing antd's layout mechanism, but figured I'd save it for a separate PR.  That said your nudge prompted me to submit so you could take a look.

The `display: grid` mechanism makes this remarkably simple. Let me know what you think of the proposed API and we can refine / chat about whether to pull the trigger on making the replacement as part of this change.

ui/src/demo/pages/Welcome.tsx
codeviking: I figured I'd let each page set it's layout, given I could foresee different pages wanting different layouts.

ui/src/demo/pages/Welcome.tsx
codeviking: Agree'd. This was just to making the prior convention (where the content on the home / welcome page is centered, and the "default" layout but the header is in the "app" layout.

I'd actually propose we just modify the home page to use the default layout across the board. It's a little odd that the header shifts, but it (in my opinion) makes the Welcome page feel a bit better.  Let's chat tomorrow about what to do here.

ui/src/demo/pages/design/Design.tsx
codeviking: I renamed them to `hcenter` and `with-left-pane` at one point. Let's come up with some names in person with @Pelagic-sky's help, as I bet he'll have some good ideas to add too.

ui/src/demo/pages/design/Design.tsx
codeviking: Sure, can change 👍

ui/src/lib/components/Footer.tsx
codeviking: Good call, can do.

ui/src/lib/components/Layout.tsx
codeviking: I don't think it's necessary anymore.  I could, however, be wrong.

codeviking:
New adhoc:

https://varnish.018939.apps.allenai.org

To preview the new layout mechanism:

https://varnish.018939.apps.allenai.org/scratch?variant=app
https://varnish.018939.apps.allenai.org/scratch?variant=default

ui/src/demo/pages/Welcome.tsx
jonborchardt: different pages will want different layouts... but not sure they will want different variants...
unless we change to something like align="left"
then i agree different pages will want different align props

ui/src/demo/pages/Welcome.tsx
jonborchardt: 2 sep issues..
1- what to do in the demo
2- what to support

1- i think we agree, that the whole demo should use 'app'
2- im not sure we should support changing layout variants on different pages, lets discuss in person

ui/src/lib/layout/index.ts
```diff
@@ -0,0 +1,4 @@
+import { createContext } from "react";
+import { LayoutVariant } from "../components/Layout";
+
+export const Layout = createContext<LayoutVariant>('default');
```
jonborchardt: rename export to LayoutContext?

ui/src/demo/pages/Scratch.tsx
```diff
@@ -0,0 +1,136 @@
...
+    display: grid;
+    min-height: 0;
+    grid-template-columns: ${({ collapsed }) => collapsed ? "5vw" : "min-content"} auto;
+`;
+
+const LeftMenu = styled.aside`
+    background: #fff;
+    margin: 0;
+    border-right: 1px solid ${({ theme }) => theme.palette.border.main};
+    min-width: 20vw;
```
jonborchardt: min-width needs to be configurable by the consuming app

ui/src/demo/pages/Scratch.tsx
```diff
@@ -0,0 +1,136 @@
...
+`;
+
+const ScrollWindow = styled.main`
+    overflow: auto;
+`;
+
+const Main = styled.main`
+    background: #fff;
+    padding: ${({ theme }) => theme.spacing.lg};
```
jonborchardt: main likely needs a max width in both variants

ui/src/demo/pages/Scratch.tsx
```diff
@@ -0,0 +1,136 @@
...
+            )}
+        </Layout>
+    );
+}
+
+const Title = styled(HeaderTitle)`
+    padding: ${({ theme }) => theme.spacing.lg} 0;
+`;
+
+const Menu = () => (
```
jonborchardt: any reason we dont keep using the ant menu?
i like it, and it is not tied to the layout stuff

ui/src/demo/pages/Scratch.tsx
```diff
@@ -0,0 +1,136 @@
...
+    const query = qs.parse(location.search.substr(1));
+    const variant = query.variant || "default";
+    return (
+        <Layout variant={variant}>
+            <Header>
+                <HeaderColumns gridTemplateColumns="auto">
+                    <Title>Test</Title>
+                </HeaderColumns>
+            </Header>
+            {variant === "app" ? (
```
jonborchardt: lets talk in person, but i think we are overloading the term variant...
in this case, this is a layout .. not a layout variant...

ie...
common layouts
-full scrolling page
-page with a left menu
-page with a top menu
-page with footer always visible
-etc

variants
- float stuff left
- float stuff center

i dont have this completely baked in my head.. but we should talk since im confused....

ui/src/demo/App.tsx
```diff
@@ -1,20 +1,44 @@
...
+            exact: true,
+            label: 'Welcome',
+            component: Welcome
+        },
+        {
+            path: '/design',
+            label: 'Design',
+            component: Design
+        },
+        {
```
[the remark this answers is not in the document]
codeviking: I'm going to fix prior to merging, per my original comment.

ui/src/demo/pages/Scratch.tsx
codeviking: Sure, can do.

ui/src/lib/layout/index.ts
codeviking: That's not canonical per the context examples I can find, so I'm hesitant to label as as such. That said I agree that the collision with `Layout` isn't ideal.

I wonder if there's a more general thing we're doing here that could inherit this name.  I'll think about it a bit, or just cave and call it `LayoutContext`.

ui/src/demo/pages/Scratch.tsx
codeviking: Currently there isn't a max-width on the application's main content. I figured, for now, I'd match what's implemented.

Is this a known issue, i.e. does @Pelagic-sky want a max-width on the application view's content too?

ui/src/demo/pages/Scratch.tsx
codeviking: For me, it feels like overkill, as it's faster for me to just write my own HTML (as I did here).

That said this was not meant to be something that sticks, just temporary code for the example / scratch page.

ui/src/demo/pages/Scratch.tsx
codeviking: Interesting. Isn't it a `type` or `variant` of a layout, that the user is invoking? Don't we use `variant` elsewhere for a similar taxonomy?

In my mind we either:

(a) Have a single `<Layout />` with `variant` to express the variations.
(b) Have a single `<Layout />` with `type` to express the variations
(c) Use different components for different variations, i.e. `<FullPageCenteredLayout />` and `<LeftPaneNavigationLayout />`.

I'm not a fan of (c), and I think we've been using `variant` with other APIs to maybe (a) is actually ok?

codeviking:
> 2- im not sure we should support changing layout variants on different pages, lets discuss in person

Unfortunately I can't reply above. I'm not sure why.

I think we should *encourage* people to use a single layout for their entire application.  That said I think we should *allow* for people to use a separate layout on different pages (or possibly on different parts of the same page).

In my mind the APIs we provide should be opinionated (encourage a given pattern that we think nets to the best UX) but flexible to accommodate lots of different use-cases. This will ensure maximum adoption in my mind.

ui/src/demo/components/Header.tsx
codeviking: Fixed.

ui/src/demo/pages/Welcome.tsx
codeviking: I think we should *allow* pages to be different layouts, but encourage otherwise. I can think of no reason to write code that prevents them from doing so.

codeviking:
@jonborchardt I made the change to a single `app` layout in the demo. The home page then looks like this:

image


Are we ok with this?

ui/src/lib/layout/index.ts
jonborchardt: LayoutParams?
LayoutData?
LayoutSettings?

codeviking: Eh, I'll just do `LayoutContext`. I don't love it, but it's specific.

ui/src/demo/pages/Scratch.tsx
codeviking: I think we landed on `(a)` in person.  We can re-assess once this is ready for another pass.

ui/src/lib/components/Layout.tsx
jonborchardt: it is :-P, but it might get fixed with other changes you make to layout

jonborchardt:
in the adhock, if i make the window smaller than the menu, the scroll scrolls the whole window including the header

jonborchardt:
re: homepage
that is fine, however, i think that we want a max width for text, even for the 'app' variant
but we can iterate

ui/src/lib/components/Layout.tsx
codeviking: Yup, hoping to reintroduce after I get the new layout implementation in.

ui/src/lib/layout/index.ts
codeviking: Fixed.

ui/src/demo/pages/Scratch.tsx
codeviking: Implemented.

ui/src/demo/pages/Welcome.tsx
codeviking: Sounds good.  At the moment I am swapping between the two, but it's only so that I can easily test both variants. I'll redact that prior to merging.

codeviking:
@jonborchardt This is ready for another pass.

ui/src/demo/App.tsx
```diff
@@ -1,20 +1,23 @@
...
 export default class App extends React.PureComponent {
     render() {
         return (
             <BrowserRouter>
                 <React.Fragment>
-                    <GlobalStyles />
-                    <ScrollToTopOnPageChange />
                     <ThemeProvider>
-                        <Route path="/" component={Home} />
+                        <GlobalStyles />
```
codeviking: Placing this as a child of `<ThemeProvider />` allows it to use our theme, which seems ideal.

ui/src/demo/pages/design/Breakpoints.tsx
```diff
@@ -65,13 +65,13 @@ export class Breakpoints extends React.PureComponent<RouteComponentProps> {
     render() {
         return (
             <React.Fragment>
-                <PageTitle>Breakpoints</PageTitle>
+                <h1>Breakpoints</h1>
```
codeviking: I removed `PageTitle`, as I think that:

* Our `<h1 />` should be sufficient.  If it's not, let's work with @Pelagic-sky to figure out what additional styles it needs.
* We were using a `<h2 />`, which was incorrect.  This seems to be the most prominent title for the page's content.

codeviking:
To do:

- [ ] Figure out why it's not rendering on mobile (Chrome, Brave or Safari all simply render a blank page).

codeviking:
Unfortunately the mobile issues make this a no-go.  Going to back this close this out.  We might implement some of the mechanics later.

PR #159, opened 2019-09-04
proposal for adding antd components

jonborchardt:
this allows us to use ant examples as is... while making it a bit harder to develop the components on our side

maybe this could be auto generated by a script?


ui/src/lib/components/Tabs.tsx
```diff
@@ -1,7 +1,26 @@
 import { Tabs as AntTabs } from 'antd';
-import styled from 'styled-components';
+import styled, { StyledComponentBase } from 'styled-components';
 
-export const TabVarnish = styled(AntTabs).attrs({ animated: false })`
+// exporting TabPane as a convenience
+// we could choose to not export, and just let users grab from Tabs below
+export const TabPane = styled(AntTabs.TabPane)``;
```
codeviking: Let's not export.  I think less is more in this case, as more just complicates things (i.e. the paradox of choice).

ui/src/lib/components/Tabs.tsx
```diff
@@ -1,7 +1,26 @@
...
+
+// Our exported styled component version of AntTabs will need to have a reference to the styled
+// component version of TabPane
+// there is a type swap here, this makes typescript happy (Antd.Tabs.TabPane => TabPane)
+// this is fine, since consumers will use the styled version and 'ref' grabs the inner component
+class VarnishTabs extends AntTabs {
+    static TabPane = TabPane as any;
+}
```
codeviking: Why is `TabPane` an `any`? Can't we use a specific type here?

ui/src/lib/components/Tabs.tsx
```diff
@@ -10,8 +29,4 @@ export const TabVarnish = styled(AntTabs).attrs({ animated: false })`
     && .ant-tabs-tabpane {
         transition: none;
     }
-`;
-
-export class Tabs extends AntTabs  {
-    static Varnish = TabVarnish;
-};
+` as any as WithStaticProps;
```
codeviking: Why do we have to cast this to both `any` and `WithStaticProps`?

ui/src/lib/components/Tabs.tsx
```diff
@@ -1,7 +1,26 @@
...
+// component version of TabPane
+// there is a type swap here, this makes typescript happy (Antd.Tabs.TabPane => TabPane)
+// this is fine, since consumers will use the styled version and 'ref' grabs the inner component
+class VarnishTabs extends AntTabs {
+    static TabPane = TabPane as any;
+}
+
+// AntTabs has static properties, but they are lost when converting to a styled component
+// this interface tells typescript about the styled component version of the statics
+interface WithStaticProps extends StyledComponentBase<typeof AntTabs, any, {}, never> {
```
codeviking: Naming is hard. Maybe `StyledWithTabPane`?

ui/src/lib/components/Tabs.tsx
```diff
@@ -1,7 +1,26 @@
...
+    TabPane: typeof TabPane;
+}
+
+// finally, we export our styled Tab, and tell typescript its a WithStaticProps
+export const Tabs = styled(VarnishTabs).attrs({ animated: false })`
```
codeviking: As noted, we should switch this to `() => ({ animated: false })` to avoid the need to update this down the road (and avoid the associated warning).

ui/src/lib/components/Tabs.tsx
jonborchardt: no, since we are hacking TS here

ui/src/lib/components/Tabs.tsx
jonborchardt: yes, its a ts hack...

ui/src/lib/components/Tabs.tsx
jonborchardt: i wanted to pick a name that could be used by multiple components, to make the pattern easier to copy/paste

ui/src/lib/components/Tabs.tsx
jonborchardt: will do

ui/src/lib/components/Tabs.tsx
jonborchardt: will remove

jonborchardt:
@codeviking  ok, take another look please

ui/src/lib/components/Tabs.tsx
codeviking: Gotcha.

I might write this as:

```
const Tabs = styled(VarnishTabs).attrs(...)`
   ...
`
// We trick Typescript here by first casting to any and then casting to the type we'd like to export.
// We do this so that the exported type includes the TabsPane property, which we lose otherwise.
export const Tabs as any as WithStaticProps;
```

Nesting the comment just above the line helps, I think.

ui/src/lib/components/Tabs.tsx
codeviking: Sure, sounds good.

ui/src/lib/components/Tabs.tsx
codeviking: Okie dokie.

codeviking:
@jonborchardt On another note, are there instances that we need to refactor elsewhere in the codebase that you're aware of? Maybe we should create an issue with an inventory of things we need to update to match this pattern?

ui/src/lib/components/Tabs.tsx
jonborchardt: you mean? 
```
// Finally, we export our styled Tab, and tell typescript its a WithStaticProps.
const StyledTabs = styled(VarnishTabs).attrs((props) => ({
    animated: false
}))`
    &,
    .ant-tabs-nav-container {
        font-size: inherit;
    }

    && .ant-tabs-tabpane {
        transition: none;
    }
`;

// We trick Typescript here by first casting to any and then casting to the type we'd like to export.
// We do this so that the exported type includes the TabsPane property, which we lose otherwise.
export const Tabs = StyledTabs as any as WithStaticProps;
```

because i cannot do the syntax you propose

jonborchardt:
@codeviking ill start a doc, but basically, all components need to convert to this... even if they dont expose statics, i think the pattern should be added for clarity....

codeviking:
>  ill start a doc, 

I think a Github issue suffices!

ui/src/lib/components/Tabs.tsx
codeviking: Ah, sure! If you like it as well.  If you prefer the existing, that's ok too.

Issue #168, opened 2019-09-06
Deduplicate Layout APIs.


codeviking:
The current layout mechanism seems like it could be simplified a bit, to make it easier to follow / use.  Specifically:

* `LayoutHeader` is confusing, how is it different than a `Header`?
* `LayoutFooter` is confusing, how is it different than a `Footer`?
* `ContentAndFooterLayout` seems like a weird name. Maybe we just use `Layout`?

PR #169, opened 2019-09-07
Fix #141; Fix a few layout and navigation related bugs.

### LINKED ISSUE #141: The left navigation menu has a gap at the top when the first item is selected.
image

codeviking:
This fixes #141 by adding the current height of the header to
`<LayoutContext />` so that other controls (like our fixed menu)
can adjust their position based upon the height of the header.

I also fixed a few other small bugs along the way:

* Ensure the top and left navigation items are active when
  directly navigating to these pages.
* Set the collapsed width to that which antd sets. It appears you
  can't set a collapsed width, or we're doing something wrong.

My largest takeaway here is that while this is in an ok state, Ant's
underlying `Menu`s aren't terribly flexible.


codeviking:
Re #168; This removes `ContentAndFooterLayout`, as now another `Layout` (akin to how antd advices) can be used instead.

ui/src/demo/Home.tsx
```diff
@@ -34,9 +38,20 @@ export default class Home extends React.PureComponent<RouteComponentProps> {
...
     render() {
         return (
-            <LayoutContext.Provider value={{ layoutVariant: "app" }}>
+            <LayoutContext.Provider value={{
+                layoutVariant: "app",
+                currentHeaderHeight: this.state.currentHeaderHeight,
+                setHeaderHeight: (currentHeaderHeight: number) => {
+                    this.setState({ currentHeaderHeight })
+                }}}
+            >
```
codeviking: Note that this currently has to be implemented by the user. I figure as we further canonicalize the layout mechanism we'll likely abstract this so people don't have to think about it.

ui/src/demo/Home.tsx
```diff
@@ -49,7 +64,7 @@ export default class Home extends React.PureComponent<RouteComponentProps> {
                             <HeaderSubTitle>{packageJson.version}</HeaderSubTitle>
                             <Spacer />
                             <TopMenu
-                                defaultSelectedKeys={[this.props.location.pathname]}>
+                                defaultSelectedKeys={[`/${this.props.location.pathname.split('/')[1]}`]}>
```
codeviking: This ensures the proper route is active when navigating directly to a page other than the index.

ui/src/lib/components/Layout.tsx
```diff
@@ -67,63 +47,47 @@ const ContentContainer = styled(AntLayout.Content)<{layout?: LayoutVariant}>`
...
-        )
+const FixedContainer = styled.div<{ paddingTop: number, width: number | string, collapsedWidth: number | string, breakpoint: string }>`
+    border-right: 1px solid ${({ theme }) => theme.palette.border.main};
+    position: fixed;
+    top: 0;
+    bottom: 0;
+    width: ${({ width }) => width};
+    padding-top: ${({ paddingTop }) => paddingTop}px;
+    transition: padding-top 200ms ease-in-out;
+    overflow: auto;
```
codeviking: Migrating `position: fixed` to a child allows us to remove `ContentAndFooterLayout`, as now the `LeftSider`'s dimensions properly cause it's sibling `Layout` to increase / decrease in size.

ui/src/demo/Home.tsx
```diff
@@ -19,7 +19,11 @@ import { ResponsiveWindowImage,
 } from '../lib/components';
 import { LayoutContext } from '../lib/layout';
 
-export default class Home extends React.PureComponent<RouteComponentProps> {
+interface State {
+    currentHeaderHeight: number;
```
jonborchardt: can we hide this from the end user by embedding it in a context?
seems harsh to make the user do this in every app

jonborchardt: i think we should hold it in layoutcontext

ui/src/demo/Home.tsx
```diff
@@ -34,9 +38,20 @@ export default class Home extends React.PureComponent<RouteComponentProps> {
...
+    constructor(props: RouteComponentProps) {
+        super(props);
+        this.state = { currentHeaderHeight: 0 };
+    }
+
     render() {
         return (
-            <LayoutContext.Provider value={{ layoutVariant: "app" }}>
+            <LayoutContext.Provider value={{
+                layoutVariant: "app",
```
jonborchardt: in general
i want to move away from any of these:
```
interface NamesToAvaoid {
  variant: string;
  type: string;
  namesToAvoidVariant: string;
}
```
because they bit us as soon as we have a multi variant component
```
interface NamesThatAreBetter {
  colorVariant: string;
  appType: string;
  sizeVariant: string;
}
```

ui/src/demo/Home.tsx
jonborchardt: ya, i think we should do this, this qtr.
before upgrading allennlp nor prior

ui/src/demo/Home.tsx
jonborchardt: candidate for a util function to make clearer for end users?

ui/src/lib/components/Layout.tsx
```diff
@@ -67,63 +47,47 @@ const ContentContainer = styled(AntLayout.Content)<{layout?: LayoutVariant}>`
...
-                {this.props.children}
-            </StyledSider>
-        )
+const FixedContainer = styled.div<{ paddingTop: number, width: number | string, collapsedWidth: number | string, breakpoint: string }>`
+    border-right: 1px solid ${({ theme }) => theme.palette.border.main};
+    position: fixed;
+    top: 0;
+    bottom: 0;
+    width: ${({ width }) => width};
+    padding-top: ${({ paddingTop }) => paddingTop}px;
```
jonborchardt: we have a util func
```padding-top: ${({ paddingTop }) => convertPixelsToRem(paddingTop);```

File: ui/src/lib/components/Menu.tsx [hunk 7]
```diff
@@ -11,30 +11,33 @@ export const TopMenu = styled(Menu).attrs({
...
     }
 `;
 
 export const MenuItem = styled(Menu.Item)``;
 
 export const TopMenuItem = styled(MenuItem)`
-    &&& {
-       
+    a:hover {
+        text-decoration: none;
```
jonborchardt: :-)

ui/src/lib/components/Menu.tsx
```diff
@@ -11,30 +11,33 @@ export const TopMenu = styled(Menu).attrs({
...
     }
 `;
 
-
-export const LeftMenu = styled(Menu).attrs({
-    mode: "inline"
-})``;
+export const LeftMenu = styled(Menu).attrs(() => ({ mode: 'inline' }))`
```
jonborchardt: i like to put attrs on new line.. makes a nice indent of the css:
```
LeftMenu = styled(Menu).attrs(() => ({ 
    mode: 'inline'
    // more attrs 
}))`
    border-right: none;
    // more css
`
```

ui/src/lib/layout/LayoutContext.tsx
```diff
@@ -27,5 +32,7 @@ interface LayoutSettings {
  * per page. More complex scenarios, however, might necessitate using multiple.
  */
 export const LayoutContext = createContext<LayoutSettings>({
-    layoutVariant: 'default'
+    layoutVariant: 'default',
+    currentHeaderHeight: 0,
```
jonborchardt: hmm... is current used?
i thought this pr left it up to the calling code

ui/src/demo/Home.tsx
codeviking: I'm not sure what to generalize (yet).  This is just ensuring we only pass in the first portion of the path (i.e. `/default` in `/default/path`), since the "keys" in the menu aren't the fully resolved path.  This seems somewhat specific to this use-case.

I think we'll abstract this if / when we write some sort of higher level navigation abstraction, does that sound good?

ui/src/demo/Home.tsx
codeviking: Sure, so is there something I need to rename here, or is this good as is? I'm unclear.

ui/src/lib/components/Layout.tsx
codeviking: In this case it's measured as pixels, and *should* be pixels (I think), since it's the result of `getBoundingClientRect().height` on another element -- so it shouldn't necessarily change depending on the user's font-size settings.

Of course, the user's font-size settings might change the height of the header, which in turn would change this value, so I think we're good?

ui/src/lib/components/Menu.tsx
codeviking: This is desired, right?

ui/src/lib/components/Menu.tsx
codeviking: Sure

ui/src/lib/layout/LayoutContext.tsx
codeviking: Let's chat in person, I don't follow.

ui/src/demo/Home.tsx
codeviking: https://github.com/allenai/varnish/issues/171

ui/src/lib/layout/LayoutContext.tsx
codeviking: Resolved by providing a default context.

ui/src/demo/Home.tsx
codeviking: After chatting in person we think this *might* become `AppSettings` and accordingly get a new name, but we're not sure yet so we'll rename in the future if warranted.

ui/src/demo/Home.tsx
codeviking: Done.

PR #174, opened 2019-09-14
Fix #172; A little documentation cleanup.

### LINKED ISSUE #172: The README is leaking internal code repository URLs.
The Varnish README currently links to private repositories.  These links end up on NPM:

https://www.npmjs.com/package/@allenai/varnish

Github's authentication prevents this from being an issue, but I think we should remove the links regardless, as it's leaking a bit of information externally that shouldn't be.

We should review the README and `package.json` to make sure there's no other internal information that's being published externally via this same mechanism.

codeviking:
Remove internal repository URLs from the README + some smallish
documentation cleanup.


Contributing.md
```diff
@@ -0,0 +1,67 @@
...
+
+   ```bash
+   npm version major|minor|patch
+   ```
+
+1. Commit and push the change, and make sure to include tags:
+
+   ```bash
+   git push --tags origin master
+   git push origin master
```
codeviking: I'm not sure if both are actually required, it might be that the explicit addition of the remote + branch resolves the issue (and I can simplify).  We'll update after giving this a shot :).

PR #182, opened 2019-09-24
update the menu

jonborchardt:
image


codeviking:
The new design looks really nice.  One nit, should the `NEW` label have some horizontal space between it and the `Tabs` text?

.gitignore
```diff
@@ -28,3 +28,5 @@ demo/build/
 # Parcel files
 .cache
 dist/
+
+.vscode
```
codeviking: I usually just add this to my global .`gitignore` file:

```bash
git config --global core.excludesfile ~/.gitignore
echo ".vscode" > ~/.gitignore
```

Given it's client specific, for purity's sake I tend to not add this to project `.gitignore` files.  I do the same for `.DS_Store` files.

I don't feel terribly strongly, so if you'd prefer to leave in that's ok too.

components/Menu.tsx
```diff
@@ -1,41 +1,31 @@
...
-        line-height: 77px;
-        border-bottom: 0;
-    }
-`;
-
+// Styled version of Menu statics.
+export const MenuSubMenu = styled(Menu.SubMenu)``;
 export const MenuItem = styled(Menu.Item)``;
+export const MenuDivider = styled(Menu.Divider)``;
+export const MenuItemGroup = styled(Menu.ItemGroup)``;
```
codeviking: Isn't this contrary to the new way we decided to do things? If we intend to make the change to the new mechanism later, that's fine.

I see we have the new mechanism below. Should we, accordingly, not `export` this so people bind themselves to the API we prefer?

components/Menu.tsx
```diff
@@ -53,3 +44,46 @@ export const IconMenuItemColumns = styled(Columns).attrs({
...
+    width: calc(100% - 1px);
+`;
+
+// Styled top version of Menu.
+const StyledTopMenu = styled(VarnishMenu).attrs({
+    mode: "horizontal"
+})`
+    && {
+        display: flex;
+        line-height: 77px;
```
codeviking: Maybe just stick a `TODO` here to get rid of the `line-height` based approach? Ideally we'd use padding instead, so the text could wrap cleanly.

components/Menu.tsx
```diff
@@ -53,3 +44,46 @@ export const IconMenuItemColumns = styled(Columns).attrs({
...
+    static Item = MenuItem as any;
+}
+
+// Styled left version of Menu.
+const StyledLeftMenu = styled(VarnishMenu).attrs(() => ({
+    mode: 'inline'
+}))`
+    border-right: none;
+    transition: none;
+    width: calc(100% - 1px);
```
codeviking: Is `calc` supported across all of the browsers we intend to support?

Couldn't we just put the `border` on this, and thereby just set `width: 100%`?

demo/src/AppRoute.ts
```diff
@@ -1,9 +1,36 @@
...
+    /* The component that's rendered when that route is active. */
     component: ComponentType<any>;
+
+    /* props to pass to component */
+    componentProps?: any;
+
+    /* gray out in menu */
+    disabled?: boolean;
+
+    tag?: string;
```
codeviking: Add a comment for consistency?

demo/src/ComponentGroups.tsx
```diff
@@ -0,0 +1,179 @@
...
+        path: '/',
+        exact: true,
+        label: 'Welcome',
+        iconSrc: paletteIcon,
+        component: About
+    }
+];
+
+const componentRoot = '/components';
+export const componentGroups: AppRouteGroup[] = [
```
codeviking: Since the export is a static object, not a class, can we name this file `lowerCase`, to match it's export?

demo/src/Home.tsx
```diff
@@ -1,68 +1,168 @@
...
+        this.setState({ menuCollapsed: !this.state.menuCollapsed });
+    }
+
     render() {
         return (
             <DefaultLayoutProvider layoutVariant="app">
                 <Layout bgcolor="white">
                     <Header>
-                        <HeaderColumns gridTemplateColumns="auto auto 1fr auto">
+                        <HeaderColumnsWithSpace gridTemplateColumns="auto auto 1fr">
```
codeviking: What's the `WithSpace` for?  Shouldn't `grid-gap` be sufficient?

demo/src/Home.tsx
```diff
@@ -1,68 +1,168 @@
...
-        );
+        )
+    }
+}
+
+const HeaderColumnsWithSpace = styled(HeaderColumns)`
+    margin: 11.5px 0;
+`;
+
+class IconMenuItem extends React.PureComponent<{ route: AppRoute, menuCollapsed: boolean }> {
```
codeviking: This seems like a good candidate for a stateless functional component?

demo/src/Home.tsx
```diff
@@ -1,68 +1,168 @@
 import * as React from 'react';
```
codeviking: Since we're doing a fair bit of restructuring in this PR, should we consider renaming this to `AppChrome` or something like that, to reflect that it's not a `Home` but rather a top-level container of some sort?

demo/src/Home.tsx
```diff
@@ -1,68 +1,168 @@
...
                 </Layout>
             </DefaultLayoutProvider>
-        );
+        )
+    }
+}
+
+const HeaderColumnsWithSpace = styled(HeaderColumns)`
+    margin: 11.5px 0;
+`;
```
codeviking: this doesn't feel quite right.  I'd expect `padding` instead of `margin` here, as we don't want this value to ever collapse.  I'd also expect the padding to be on the parent, and/or possibly each column.  I'd propose thinking about this one a bit more (but don't know the right answer).

.gitignore
jonborchardt: ill remove

components/Menu.tsx
jonborchardt: yes, oops.. they need to exist, but not be exported

components/Menu.tsx
jonborchardt: k

demo/src/AppRoute.ts
jonborchardt: will do

components/Menu.tsx
jonborchardt: just make border transparent?

demo/src/ComponentGroups.tsx
jonborchardt: frankly not sure how to do that without changing the file name to something different... the whole OSX/github not understanding case changes

demo/src/Home.tsx
jonborchardt: this is space above and below the grid

demo/src/Home.tsx
jonborchardt: ill convert

demo/src/Home.tsx
jonborchardt: sure

demo/src/Home.tsx
jonborchardt: ill convert

components/Menu.tsx
jonborchardt: that worked

PR #193, opened 2019-10-10
Use the shared eslint configuration.

codeviking:
This removes the `eslint` configuration, and uses that published
via `@allenai/eslint-config-varnish` instead.

As part of this I also moved all `eslint` config out of `package.json`. 
I tend to avoid using `package.json` to configure anything other
than the dependencies that are installed, as I find it doesn't scale
well over time (the `package.json` file becomes really hard to read,
edit, etc).


package.json
```diff
@@ -18,85 +18,6 @@
...
-      "ecmaVersion": 2018,
-      "sourceType": "module"
-    },
-    "plugins": [
-      "react",
-      "@typescript-eslint",
-      "prettier"
-    ],
-    "rules": {
-      "prettier/prettier": [
```
jonborchardt: we want to keep these rule overrides

jonborchardt: n/m i get it

codeviking: They're in the shared config now?

Issue #176, opened 2019-09-18
Use AI2's canonical name in all mentions.


codeviking:
We should update references to AI2 (by name) to be consistent and use the canonical form.  We should use "Allen Institute for AI" at all times, instead of "Allen Institute for Artificial Intelligence", "Ai2", etc.

I think this is only in the Footer, but we should make sure any other references are updated.

aaronsarnat:
@codeviking is this a direct quote from Oren?

> We should update references to AI2 (by name) to be consistent and use the canonical form. We should use "Allen Institute for AI" at all times, instead of "Allen Institute for Artificial Intelligence", "Ai2", etc.

codeviking:
No, but it was the intent.

aaronsarnat:
Cool. I was writing an email to Jen and Carissa re: legal mentions but ended up paraphrasing the same intent since i wasn't sure

codeviking:
Ah, I'd just emphasize that we're consolidating verbiage from a branding perspective.  I'm not sure what our legal / incorporated name is -- that's likely what should go in those docs.

codeviking:
Once complete, let's go through the same process as https://github.com/allenai/varnish/issues/205 (just for the crucial, must have ones) to make sure the verbiage is consistent.

codeviking:
@aaronsarnat If you haven't heard back from Jenn by tomorrow, let's just drop by her office and ask her in person. I bet she could answer (regarding just the copyright) very quickly. The privacy policy and terms of use might be a trickier aspect of this that may require more time.

codeviking:
I believe this is now resolved.  The name in the Footer shouldn't change, as per legal it needs to be our "Legal" name not our "Doing Business Name."

PR #209, opened 2019-11-08
Re #176; Update AI2 name in footer.

codeviking:
Use the canonical form of AI2's name in the
`<Footer />`.


codeviking:
We should likely wait to merge this until legal approves it, @aaronsarnat has reached out.

aaronsarnat:
Yeah, as @codeviking mentioned, it's possible we can't do this from a legal perspective. Either way, we probably want to get rid of the `(AI2)` text.

I will reach out to Jen again today to hopefully get final word on org name policy in copyright line.

components/Footer.tsx
```diff
@@ -52,9 +52,9 @@ export class Footer extends React.PureComponent<Props> {
                             this.props.children
                         ) : (
                             <Body>
-                                © The{' '}
+                                ©{' '}
                                 <ExternalLink contrast={contrast} href="https://allenai.org">
-                                    Allen Institute for Artificial Intelligence (AI2)
+                                    Allen Institute for AI
```
aaronsarnat: Got final word from Jen. This line is considered legal copy, and as such, needs to reference our legal entity name, not our friendly DBA name. So this reference should be:

```
The Allen Institute for Artificial Intelligence
```

Which includes "The" at the beginning, does not use the "AI" acronym, and does not include "(AI2)."

codeviking: 🤘 Updated, please confirm.

codeviking:
@aaronsarnat Take another look, and let me know if it's good to go!

PR #210, opened 2019-11-08
Adding AI2logo component

aaronsarnat:
image

**This PR:**

- Adds a new component called `<AI2Logo />`
- Incorporates svg assets that represent the different logo color and layout variants
- Adds a page for usage documentation in Varnish demo under "General" section
- Updates `<AI2Banner />` component to use new `<AI2Logo />` component
- Updates favicon


aaronsarnat:
In addition to review, I could definitely benefit from having one of you guys walk me through the process of releasing a new version. Jon and I did this several weeks ago, but I could use a refresher course.

components/AI2Logo.tsx
```diff
@@ -0,0 +1,71 @@
...
+                break;
+            case 'lg':
+                imgHeight = 37;
+                imgWidth = includeText ? 326 : 78;
+                break;
+            default:
+                imgHeight = 30;
+                imgWidth = includeText ? 264 : 64;
+        }
+    }
```
codeviking: I'd consider factoring these out into functions, the benefit being:

1. The mutable variables can be isolated to a smaller scope.
2. I think it makes the `render()` easier to read in entirety, and lets the user choose where to dive in.

So, I'd for instance write two methods:

```
function getImageSrc(includeText: bool, size: LogoSize, color: LogoColor): string {}
function getImageDimensions(size: LogoSize): Dimensions {}
```

Then in the render block you'd do:

```
const { width, height } = getImageSize(...)
return <VectorImage src={getImageSrc(...)} width={width} height={height} alt="..." />
```

Let me know what you think.

aaronsarnat: @codeviking just to clarify, the `...` is `code goes here` and not spread operators, right? The way I interpreted your suggestion is that the final render would look like the following. Did I understand correctly?

```jsx
const { width, height } = getImageSize(includeText, size);

return <VectorImage
         src={getImageSrc(color, includeText, size)}
         width={width}
         height={height}
         alt="Allen Institute for AI logo" />
```

So basically, the render is passing the component props to those get functions you proposed as arguments?

aaronsarnat: Also, how would you imagine the proposed `Dimensions` type would be interfaced given the destructuring? I've never tried doing this before...

`const { width, height } = getImageSize(includeText, size);`

Thanks for the feedback!

components/AI2Logo.tsx
```diff
@@ -0,0 +1,71 @@
...
+        <VectorImage
+            src={imgSrc}
+            width={imgWidth}
+            height={imgHeight}
+            alt="Allen Institute for AI logo"
+        />
+    );
+};
+
+const VectorImage = styled.img`
```
jonborchardt: can we move this to util and export it?

jonborchardt:
this is fine to ship, but im wondering if we can make it more generic...
supply a list of images, a list of sizes, and you get all the variant stuff for 'free'...
just a thought for later.. aka, as soon as we do this to a second image.... :-P

components/AI2Logo.tsx
codeviking: Yup, you've got it.

Assignment via destructuring will work as expected. Give it a shot and shoot me a message on Slack if it doesn't work.

components/AI2Logo.tsx
aaronsarnat: @jonborchardt I'm not following. Why make this a util?

codeviking: He's just proposing that it becomes a more generic, re-usable thing (i.e. so other components can use it).  Skip it for now, we can generalize down the road.

aaronsarnat: Gotcha.

aaronsarnat: @jonborchardt I cut an issue that captures your feedback about considering generalizing some of this component at some point. https://github.com/allenai/varnish/issues/211

codeviking: Stray thought -- I'd vote we use `TODO`s instead of Github issues.  Github issues can very quickly get too a point where there's so much noise it's untenable.  TODOs are much less invasive, but can still be easily found.  Furthermore they provide prompts inline in the source code (so that when someone reads / touches this, they revisit that point).

I'd keep issues a canonical place to capture real bugs and/or features that are immediately actionable.  Potential refactors / improvements in my mind aren't that.

Let me know what you two think.

aaronsarnat: that's fine, i can make this a todo

components/AI2Logo.tsx
aaronsarnat: Cool, I think I got it working. This is a neat trick. Good call!

aaronsarnat:
@codeviking PTAL

components/AI2Logo.tsx
aaronsarnat: One detail: I converted my switch statement into an if/then/else block because eslint was complaining about unreachable code. As of my latest commit, eslint is happy and I verified that the updated logic renders as intended.

components/AI2Logo.tsx
```diff
@@ -0,0 +1,87 @@
...
+            return color === 'white' ? logoWhite : logoFullColor;
+        }
+    }
+
+    function getImageDimensions(includeText: boolean, size: LogoSize | number): Dimensions {
+        if (typeof size === 'number') {
+            const defaultAspectRatio: number = 44 / 5;
+            const noTextAspectRatio: number = 78 / 37;
+            return {
+                width: Math.round(size * (includeText ? defaultAspectRatio : noTextAspectRatio)),
```
aaronsarnat: @codeviking what's better here in your opinion? fewer lines (e.g. putting the aspect ratios inline here) or readability? (keeping them as consts for better context)

codeviking: Always err towards readability over brevity.  Brevity is only helpful when it aids readability.  Brevity is bad if it hampers readability.

This method after being reimplemented is a tad hard to read. Proposal for readability incoming shortly.

components/AI2Logo.tsx
```diff
@@ -0,0 +1,87 @@
...
+            }
+        } else {
+            return color === 'white' ? logoWhite : logoFullColor;
+        }
+    }
+
+    function getImageDimensions(includeText: boolean, size: LogoSize | number): Dimensions {
+        if (typeof size === 'number') {
+            const defaultAspectRatio: number = 44 / 5;
+            const noTextAspectRatio: number = 78 / 37;
```
codeviking: You don't need `:number` here, as the type is implicit.

components/AI2Logo.tsx
```diff
@@ -0,0 +1,87 @@
...
+                return size === 'micro' ? logoWhiteWithTextMicro : logoWhiteWithText;
+            } else {
+                return size === 'micro' ? logoFullColorWithTextMicro : logoFullColorWithText;
+            }
+        } else {
+            return color === 'white' ? logoWhite : logoFullColor;
+        }
+    }
+
+    function getImageDimensions(includeText: boolean, size: LogoSize | number): Dimensions {
```
codeviking: I'd swap the parameter order. Important params first.  That way you could also set a default for `includeText`, simplifying it's call signature:

```
getImageDimensions(size: LogoSize | number, includeText: boolean = true): Dimensions
```

components/AI2Logo.tsx
codeviking: Also, don't `round`, let the browser round for us.  Use as much precision as possible, for the best image.

components/AI2Logo.tsx
codeviking: I'd break things up a bit for readability:

```TypeScript
const dimsByLogoSize: { [k: LogoSize]: Dimensions } = {
   'default': { width: 100, height: 100 },
   'sm': { width: 10 height: 10 },
   ...
}
const noTextDimsByLogoSize: { [k: LogoSize]: Dimensions } = {
   'default': { width: 6, height: 100 },
   'sm': { width: 1 height: 10 },
   ...
}
function getDimensionsFromHeight(height: number, includeText: boolean = true) {
     const byLogoSize = includeText ? dimsByLogoSize : noTextDimsByLogoSize;
     const dims = byLogoSize["default"]
     const aspectRatio = dims.width / dims.height;
     return { width: aspectRatio * height, height };
}
function getImageDimensions(size: LogoSize | number, includeText: boolean = true): Dimensions {
    if (typeof size == 'number') {
         return getDimensionsFromHeight(size);
    } else {
        const byLogoSize = includeText ? dimsByLogoSize : noTextDimsByLogoSize;
        return byLogoSize[size];
    }
}
```

I think this is a bit easier to follow.  The key is splitting out functionality into small, easy to understand pieces.

Let me know if any of it doesn't work on Slack and I can help walk through specific gotchas.  Coding in Github is always a tad tricky :).

components/AI2Logo.tsx
aaronsarnat: I figured since these are SVGs that are incapable of stretching and that don't have backgrounds, we could just set a hard px value as opposed to a floating point, but I don't have a strong feeling about it. It's going to round either way, but if you'd rather the math not happen in js then i can take out the round.

components/AI2Logo.tsx
aaronsarnat: I'm currently setting a default value of of `includeText` in the destructured props at the top of the component export. Would it be better to do it here instead? (or in addition to?)

components/AI2Logo.tsx
codeviking: > I figured since these are SVGs that are incapable of stretching and that don't have backgrounds, we could just set a hard px value as opposed to a floating point

What's the advantage of a rounded value? I don't follow.

I think without the `round` it's simpler. It's also in theory a bit faster (less work at runtime).  I also prefer to let the browser do the rounding (it will), so it's as precise as possible.

If there's a case to round, we should -- I'm just not sure if that's the case.

components/AI2Logo.tsx
codeviking: Ah, that makes sense.  No, as it's a local helper, so you don't need to set it here as well.

I would however swap the parameter order, as `size` is definitely more important then `includeText`, imo.

components/AI2Logo.tsx
aaronsarnat: nah, no strong case. It's just the designer in me liking integer px values :)
but like I said... I don't have a problem removing the round.

components/AI2Logo.tsx
aaronsarnat: done

components/AI2Logo.tsx
aaronsarnat: done

components/AI2Logo.tsx
aaronsarnat: done

components/AI2Logo.tsx
aaronsarnat: I see, so basically taking the magic numbers out of the conditional logic and moving them to a lookup table?

codeviking: Yup, this:

1. Allows it to be re-used for calculating the aspect ratio (rater than hard-coding those dimensions)
2. Makes the implementation a bit more readable, I think. I this case I think the brevity where it's used adds clarity -- that said this is admittedly subjective.
3. Ports the code for calculating the height based upon the aspect ratio to a separate function, so that the purpose code for doing so can be clarified by the function name, and so that the parent function's variadic behavior is easy to follow.

aaronsarnat: @codeviking PTAL.

I used your suggestion, but tweaked it a bit so there's only 1 lookup table. I started with 2, [per your suggestion](https://github.com/allenai/varnish/pull/210/commits/d58f0598c98086da406315e7be7b694a9e8036f6), but I didn't like that the height values were repeated in both tables. I also followed the linter's advice and was able to get rid of some type detail in the lookup table obj const.

codeviking: Ah.

The functions right now are going to be defined with every `render`, as will the map.  While this likely isn't much of a performance tax in reality, it's not consistent with code elsewhere, and I think it makes the `render()` harder to read.

I think you should move the functions outside of the component scope. This will necessitate duplicating the maps, or modeling the `includeText` property into the map somehow.  I'd use 2 maps, as a bit of duplication is worth simplicity in this case.

aaronsarnat:
@codeviking I moved the functions out of render and reverted to two map consts. In an effort to make the maps more DRY, I set the height values in the second map to equal those of the first. PTAL

components/AI2Logo.tsx
```diff
@@ -0,0 +1,103 @@
...
+    },
+    default: {
+        width: 64,
+        height: dimsByLogoSize.default.height
+    },
+    lg: {
+        width: 78,
+        height: dimsByLogoSize.lg.height
+    }
+};
```
codeviking: 👍

aaronsarnat:
Merging on green

PR #213, opened 2019-11-13
Fixing AI2Logo import that was causing react-router regression

aaronsarnat:
This PR attempts to fix the issue that was uncovered in https://github.com/allenai/ai2-web/pull/58


aaronsarnat:
Thanks for identifying the what and suggesting process of elimination! I just identified the who! :D

Merging on green.

PR #216, opened 2019-11-14
allows the use of allowClear attribute

jonborchardt:


components/form.tsx
```diff
@@ -11,9 +11,20 @@ export const TextArea = styled(Antd.Input.TextArea)`
...
+        font-size: ${({ theme }) => theme.typography.bodySmall.fontSize};
+
+        /* when the user uses allowClear attribute, the control swaps from an input to a span with
+        an input in it.  so we need specific crappy css for this case */
+        input {
+            margin: ${({ theme }) => `-${theme.spacing.xs} -${theme.spacing.md}`};
+            padding: ${({ theme }) => `${theme.spacing.xs} ${theme.spacing.md}`};
+            height: initial;
+            font-size: ${({ theme }) => theme.typography.bodySmall.fontSize};
+        }
```
codeviking: Don't we need to omit the `padding` and `font-size` on the top level container when this is true (i.e. won't this create padding on both the `span` and the `input`?  Is that desired?

demo/src/pages/Forms.tsx
```diff
@@ -121,17 +121,6 @@ export class Forms extends React.PureComponent<RouteComponentProps> {
...
-                    <h4>Input</h4>
-                    <DefaultLiveProvider code={examples.input} />
-                    <br />
-                    For more information see the:{' '}
-                    <ExternalLink
-                        target="_blank"
-                        href="https://ant.design/components/input/#Input.TextArea">
-                        Ant Design Component
-                    </ExternalLink>
-                </SectionWithDivider>
```
codeviking: Why are we removing this?

components/form.tsx
codeviking: I see now, we're negating the outer margin.

Can't we detect the `allowClear` property, and just modify the style instead? Negating it seems quite tricky.

I'm thinking something like:

```
({ allowClear }) => (allowClear ? // css for child input : // regular css
```

You could re-use the css, and just scope the selector it's applied to based on the property, right?

jonborchardt: interesting.. ill give it a go

codeviking: Give it a shot. If it's a pain, and/or doesn't work, we can stick with this.

demo/src/pages/Forms.tsx
jonborchardt: its duplicated

codeviking: 🤘

components/form.tsx
```diff
@@ -10,10 +10,20 @@ export const TextArea = styled(Antd.Input.TextArea)`
     font-size: ${({ theme }) => theme.typography.bodySmall.fontSize};
 `;
 
-export const Input = styled(Antd.Input)`
-    padding: ${({ theme }) => `${theme.spacing.xs} ${theme.spacing.md}`};
-    height: initial;
-    font-size: ${({ theme }) => theme.typography.bodySmall.fontSize};
+export const Input = styled(Antd.Input)<{allowClear?: boolean}>`
+    &&& {
+        padding: ${({ theme, allowClear }) => (allowClear ? undefined : `${theme.spacing.xs} ${theme.spacing.md}`)};
```
codeviking: I think `undefined` should be `''` (empty string), otherwise you might get `undefined` in your css?

jonborchardt: this is done all over, i believe it works....

codeviking: I wonder if it "works" because `undefined` isn't valid css, so the browser just ditches it.  Browsers are very smart (and tolerant of mistakes) these days :).

The mechanism of styled components is just a template string.  Interpolating `undefined` into a template string evaluates to the string `"undefined"`:

```
~ node
> `this is ${undefined}`
'this is undefined
```

So I'd think it's the same with styled-components?

jonborchardt: when undefined, it does not exist...
image

when not undefined:
image

is there any other test i should do?

codeviking: Ah! Great! `styled-components` is doing something fancy / handling this case for us, which is great :).

Thanks for verifying!

PR #217, opened 2019-11-14
Jon/toggle button

jonborchardt:


components/form.tsx
```diff
@@ -58,4 +58,7 @@ export const RadioGroup = styled(Antd.Radio.Group)<{ vertical: boolean }>`
     ${Radio} {
         display: ${({ vertical }) => (vertical ? 'block' : 'inline-block')};
     }
+    .ant-radio-button-wrapper {
+        padding: ${({ theme }) => `0 ${theme.spacing.md}`};
+    }
```
codeviking: What if the user uses a `<RadioButton />` outside of a form? The associated css won't be included, right?

Maybe this is something we need to set in our `theme.css` file?

jonborchardt: this has no form involved...
(bad file name)
this is needed everywhere we have a .ant-radio-button-wrapper in a radiogroup
it fixes an unexplained byproduct of varnish

codeviking: Gotcha.

PR #219, opened 2019-11-14
apply the static pattern to form components

jonborchardt:


components/Input.tsx
```diff
@@ -0,0 +1,53 @@
...
+    static Search = StyledSearch as any;
+    static TextArea = StyledTextArea as any;
+    static Password = StyledPassword as any;
+}
+
+// Styled version.
+const StyledInput = styled(VarnishInput)<{ allowClear?: boolean, prefix?: any }>`
+    &&& {
+        padding: ${({ theme, allowClear, prefix }) =>
+            (allowClear || prefix) ? undefined : `${theme.spacing.xs} ${theme.spacing.md}`};
```
codeviking: I realize we confirmed `undefined` is handled properly, but I'd still consider switching to `null` since it's React's canonical suggestion: https://reactjs.org/docs/conditional-rendering.html#preventing-component-from-rendering

jonborchardt: sounds good

PR #223, opened 2019-11-18
Adding className prop to AI2Logo

aaronsarnat:
This is an attempt to fix a bug in `ai2-web` where the `<AI2Logo />` component wasn't accepting additional styles.

Thanks to @jonborchardt and @codeviking who both independently suggested this fix. :)


aaronsarnat:
Merging on green

jonborchardt:
but who was first? :-P

PR #224, opened 2019-11-18
Adding lint:fix to docs

aaronsarnat:
I forget to do this sometimes, and also forget the precise syntax for the formatter command for this project, so I thought this might be helpful.


aaronsarnat:
Merging on green

Issue #226, opened 2019-11-20
Consider adding a spacing value equal to 3rem (48px)


aaronsarnat:
I think we're missing a spacing value between `xl` and `xxl`. I suspect this might be used in multiple places on `ai2-web`.
![image](https://user-images.githubusercontent.com/8367927/69265652-767e6080-0b7e-11ea-9b65-a8792e14a0f5.png)

aaronsarnat:
@codeviking @jonborchardt @Pelagic-sky any objections to adding this to the theme?

If not, the only question is what do we call it?

Pelagic-sky:
We could call it XXL and then upgrade 64 to XXXL

aaronsarnat:
Fine with me, but as discussed in person with Nick, any users who are currently leveraging `xxl` in their apps will have those sizes decrease when they update Varnish.

@codeviking @jonborchardt any objections with this? If not, I'll go ahead and make a PR.

codeviking:
👍

PR #228, opened 2019-11-20
Adding a new spacing value between 36 and 64

aaronsarnat:
Varnish, now with `xxxl` action.

Fixes https://github.com/allenai/varnish/issues/226

---

Screenshot from demo. See `xxl` and `xxxl`:

image


demo/src/shared/PageTitle.tsx
```diff
@@ -1,5 +1,5 @@
 import styled from 'styled-components';
 
 export const PageTitle = styled.h2`
-    margin-bottom: ${({ theme }) => theme.spacing.xxl};
+    margin-bottom: ${({ theme }) => theme.spacing.xxxl};
```
codeviking: Is @Pelagic-sky ok with the additional space for page titles?

aaronsarnat: @codeviking I changed this to match what was there before. `xxxl` is the new `xxl`. Bother are 64px.

The real new value was shoehorned between 36 and 64.

Visually, this is no different than before. But as we discussed in https://github.com/allenai/varnish/issues/226 this will mean that any app that was using `xxl` before will have that spacing value shrink by 16px if they update to the latest Varnish, lest they replace instances of `xxl` with `xxxl`.

codeviking: Ah, thanks. I didn't follow. LGTM.

aaronsarnat: No worries. Thanks for the quick review.

aaronsarnat:
Merging on green

Issue #222, opened 2019-11-18
Remove `Raleway` font from library


aaronsarnat:
This was a request from @Pelagic-sky:
Remove `Raleway` so that all headings and body text are using `Lato`.

Pelagic-sky:
Headline 1: Lato Heavy 46
Headline 2: Lato Heavy 36
Headline 3: Lato Heavy 30
Headline 4: Lato Heavy 24
Headline 5: Lato Bold 18
Headline 6: Lato Medium 13

Pelagic-sky:
Same line-spacing as before.

aaronsarnat:
@Pelagic-sky we don't currently have a `Heavy` font weight in Varnish. I can add one, but the closest thing the [Google Font](https://fonts.google.com/specimen/Lato) gives us is `Black / 900`. Is that what you want for H1-4?

Here's a chart of the font weights we have access to:

![image](https://user-images.githubusercontent.com/8367927/69663927-2864d380-103c-11ea-8c14-3938211a315b.png)

Pelagic-sky:
Black is fine

PR #233, opened 2019-11-26
Removing Raleway font (and switching headings to Lato)

aaronsarnat:
Per @Pelagic-sky's request, this PR fixes https://github.com/allenai/varnish/issues/222


theme/theme.less.ts
```diff
@@ -29,7 +29,7 @@ export const less = `
      **/
 
     // the fonts we use
-    @import (css) url("https://fonts.googleapis.com/css?family=Raleway:400,600,700|Lato:400,700|Volkhov:400,700|RobotoMono:400,700");
+    @import (css) url("https://fonts.googleapis.com/css?family=Lato:400,700|Volkhov:400,700|RobotoMono:400,700");
```
jonborchardt: i think we need lato 600 now

aaronsarnat: I don't think so... nothing was actually using it it doesn't look like.

image

Also, Google's [Lato](https://fonts.google.com/specimen/Lato) doesn't give us a 600.

We do need a Heavy / 900 though which I added in my latest commit.

aaronsarnat:
Merging on green

PR #237, opened 2019-12-09
Style the InputNumber ant component so the padding is correct.

meslater1030:
This PR adds a very basic InputNumber to Varnish so the padding is consistent with other Ant inputs.


components/InputNumber.tsx
```diff
@@ -0,0 +1,21 @@
...
+const StyledAntInputNumber = styled(AntInputNumber)`
+  && {
+    padding: 0;
+    height: initial;
+    font-size: ${({ theme }) => theme.typography.bodySmall.fontSize};
+    input.ant-input-number-input {
+      padding: ${({ theme }) => `${theme.spacing.xs} ${theme.spacing.md}`};
+    }
+  }
+`;
```
codeviking: I think you'll want to follow the (nuansced) syntax here, for exporting the styled component, but with the original `antd` type information:
https://github.com/allenai/varnish/blob/master/components/Tabs.tsx#L37

I *think* that prevents type information from being lost.

@jonborchardt Would know better (and I'm going to let him handle approval, as I'm not very up to date on the latest n' greatest in Varnish).

jonborchardt: Sams comment is for components that export statics
This component has no statics...

can we just export the style?
```
export const InputNumber = styled(AntInputNumber)`
  && {
    padding: 0;
    height: initial;
    font-size: ${({ theme }) => theme.typography.bodySmall.fontSize};
    input.ant-input-number-input {
      padding: ${({ theme }) => `${theme.spacing.xs} ${theme.spacing.md}`};
    }
  }
`;
```

as is is ok, but it does add a new div around the input... i think

PR #238, opened 2019-12-09
Add components for AI2 project icons.

codeviking:
This adds components that render individual AI2 project icons
as SVG.

I added the `ai2/projects` directory in `components/`, given the
component collection is getting large. I also added `TODO`s to
move `AI2Logo` and `AI2Banner` into `ai2/` for consistency -- that
said doing so now would make this non-backwards compatible and only
necessitate a bunch of work for our users.


components/button.tsx
```diff
@@ -136,22 +105,14 @@ export const Button = styled(AntButton).attrs<{ type?: string }>((props: Compone
...
             svg,
             path {
                 color: ${props =>
                     toHexIfDefined(
-                        valueOrDefault(
-                            props,
-                            b => b.disabled.color,
-                            b => b.disabled.contrastColor
-                        )
+                        valueOrDefault(props, b => b.disabled.color, b => b.disabled.contrastColor)
```
codeviking: The changes in this file all resulted from `yarn lint:fix`.

components/ai2/projects/icon.tsx
```diff
@@ -0,0 +1,185 @@
+import React from 'react';
+
+interface IconProps {
+    className?: string;
+    width?: string;
+    height?: string;
+}
+
+const defaults = {
+    SIZE: 32,
```
jonborchardt: get from defaulttheme?

components/ai2/projects/icon.tsx
```diff
@@ -0,0 +1,185 @@
+import React from 'react';
+
+interface IconProps {
```
jonborchardt: this interface belongs in a more general location

components/ai2/projects/icon.tsx
```diff
@@ -0,0 +1,185 @@
...
+    className?: string;
+    width?: string;
+    height?: string;
+}
+
+const defaults = {
```
jonborchardt: as do these defaults

components/ai2/projects/icon.tsx
codeviking: I don't disagree, but I'll let that generalization follow later when / if.

components/ai2/projects/icon.tsx
codeviking: In this case I'm not sure.

The icons are, natively, rendered at 32 x 32. So this feels like a more specific thing. I'm not sure though. We can always re-evaluate down the road.

components/ai2/projects/icon.tsx
codeviking: ☝️

PR #239, opened 2019-12-11
Updating bodyJumbo line-height

aaronsarnat:
Per Nick's design


aaronsarnat:
Trivial change. Merging.

PR #244, opened 2019-12-14
Ensure the modal appears on top of the header

meslater1030:
The Varnish modal was disappearing under the new header. This fixes that.


theme/zIndex.ts
```diff
@@ -23,8 +23,8 @@ export const zIndex = {
     badge: 10,
     pickerPanel: 10,
     popupClose: 10,
-    modal: 1000,
-    modalMask: 1000,
+    modal: 1080,
```
jonborchardt: can you sort these by number.... ie, move these 2 lines down

PR #246, opened 2019-12-18
Updating varnish heading color

aaronsarnat:
This PR addresses https://github.com/allenai/ai2-web/issues/194
Will fix as soon as varnish is updated in ai2-web

Solved by adding an additional value to `palette.text`

image


aaronsarnat:
Merging on green

PR #261, opened 2020-01-17
moving up some common css to varnish global

jonborchardt:
this will go in tandum with upgrading ai2-web


aaronsarnat:
@jonborchardt code looks fine, but I'm a bit scared about unforeseen side effects this could introduce on ai2-web. That said, if we merge this into Varnish then we can test before/after when we pull latest into ai2-web and make a determination at that point. My 2¢ is that we should scrutinize the home and careers pages in a branch pulling latest varnish compared to current version of varnish.

theme/theme.less.ts
```diff
@@ -89,8 +89,52 @@ export const less = `
...
+        text-decoration: ${link.hover.decoration};
+    }
+
+    a[href]:active {
+        color: ${link.activeColor};
+    }
+
+    a.ant-btn {
+        line-height: inherit;
+    }
```
codeviking: What if we group this one separately from the others, and add comments explaining each:

```
/** These are global styles we apply to all things that antd doesn't, to make it easier to 
      use raw css + html */

/** These are modifications to antd classnames that we apply globally to work around weird
      things they do in their styles. */
```

The `.ant-btn` one worries me a bit.  I feel like it should just be in `Button.tsx`. Why is it here, instead of doing it that way (I think that's what we're doing for other components too).

jonborchardt: i moved that oneoff

PR #264, opened 2020-01-22
Upgrading breakpoints and Icon

aaronsarnat:
This PR:

- Extends breakpoints to support `min-width` as well as `max-width`
- Adds accessibility attributes to `ImgIcon`


theme/breakpoints.ts
```diff
@@ -1,10 +1,36 @@
 import { Breakpoint } from './spacing';
 
+export enum Breakpoints {
```
aaronsarnat: Note about naming. There's already an export called `Breakpoint` (singular) so I made this one plural. It matches the plural object name we use in theme. E.g.

```
${({ theme }) => theme.breakpoints.lg}
```
```
${below(Breakpoints.LG)}
```

theme/breakpoints.ts
```diff
@@ -1,10 +1,36 @@
...
-    md: Breakpoint.fromPixels('md', 768),
-    lg: Breakpoint.fromPixels('lg', 992),
-    xl: Breakpoint.fromPixels('xl', 1200),
-    xxl: Breakpoint.fromPixels('xxl', 1600)
+    xs: Breakpoint.fromPixels('xs', Breakpoints.XS),
+    sm: Breakpoint.fromPixels('sm', Breakpoints.SM),
+    md: Breakpoint.fromPixels('md', Breakpoints.MD),
+    lg: Breakpoint.fromPixels('lg', Breakpoints.LG),
+    xl: Breakpoint.fromPixels('xl', Breakpoints.XL),
+    xxl: Breakpoint.fromPixels('xxl', Breakpoints.XXL)
```
aaronsarnat: As stated, I left this object in for backwards compatibility. We can use the new `above` and `below` functions, but the old way of doing media queries will still work.

components/icon.tsx
```diff
@@ -4,11 +4,20 @@ import { IconProps } from 'antd/lib/icon';
...
 export class ImgIcon extends React.PureComponent<ImgIconProps & IconProps> {
     render() {
-        const { src, ...other } = this.props;
-        return <Icon {...other} component={() => <img src={src} />} />;
+        const { src, alt, ariaHidden, ...other } = this.props;
+        return (
+            <Icon
+                {...other}
+                component={() => (
+                    <img src={src} alt={alt || ''} aria-hidden={ariaHidden ? 'true' : 'false'} />
```
jonborchardt: do we need the ` || ''`?

theme/breakpoints.ts
jonborchardt: we dont need this new enum
see below

theme/breakpoints.ts
```diff
@@ -1,10 +1,36 @@
...
+    LG = 992,
+    XL = 1200,
+    XXL = 1600
+}
+
+/* Usage:
+    @media ${above(Breakpoints.LG)} {...}
+*/
+export function above(breakpoint: Breakpoints) {
+    return `(min-width: ${breakpoint + 1}px)`;
```
jonborchardt: so that we can continue to use the old Breakpoint
return `(min-width: ${breakpoint.getPxValue() + 1}px)`;

components/icon.tsx
aaronsarnat: Sometimes it make sense to leave `alt` attributes blank. E.g.

`` is totally W3C valid and is preferred in some cases for accessibility.

Like you wouldn't want to have this which is considered redundant:

```
<h3>Gaggle of kittens</h3>
<img src="kittens.jpg" alt="Gaggle of kittens" />
```

Instead, you'd want this:

```
<h3>Gaggle of kittens</h3>
<img src="kittens.jpg" alt="" />
```

That said, we could do this if you'd prefer (set default value for alt and pass it down):

image

jonborchardt: ok

aaronsarnat: done

theme/breakpoints.ts
aaronsarnat: good call. done

theme/breakpoints.ts
aaronsarnat: done

aaronsarnat:
@jonborchardt feedback addressed. PTAL thanks!

aaronsarnat:
Merging on green

PR #265, opened 2020-01-23
Making Quote text normal fontweight instead of bold

aaronsarnat:


aaronsarnat:
@jonborchardt one question i have... with the mobile properties I added to the typography objects, what happens when you invoke them as theme objects like this?

```
const Component = styled.div`
    ${({ theme }) => theme.typography.bodyJumbo}
    ...
`;
```

jonborchardt:
what is wrong with changing fontsize % 
this seems odd that we only do this for 2 components

aaronsarnat:
@jonborchardt we're not using the global font-size media query reset on ai2-web. We overrid that when we started the project (after we debated that 87.5% value as a group). Basically we don't want our UI to shrink unilaterally. 

I added media queries to these 2 because these are the only two typography components we're using in ai2-web that nick asked us to change.

But if this feels weird I can just make the changes locally in ai2-web

components/typography.tsx
```diff
@@ -7,6 +8,11 @@ export const BodyJumbo = styled.span`
     font-size: ${({ theme }) => theme.typography.bodyJumbo.fontSize};
     line-height: ${({ theme }) => theme.typography.bodyJumbo.lineHeight};
     font-weight: ${({ theme }) => theme.typography.bodyJumbo.fontWeight};
+
+    @media ${({ theme }) => below(theme.breakpoints.sm)} {
+        font-size: ${({ theme }) => theme.typography.bodyJumbo.fontSizetMobile};
```
codeviking: Is this right? Is the property called `fontSizet` and not `fontSize`?

aaronsarnat: no that was a typo. good catch, thx!

aaronsarnat:
@jonborchardt @codeviking do we want to merge this? I'm ok with not merging and addressing the issue locally on ai2-web. There seemed to be some hesitation about this fix, so just want to double check with you both

jonborchardt:
i dont think we should... but if @codeviking or @Pelagic-sky really want it, i can bend...

i think that this should just be in ai2-web for now

aaronsarnat:
@jonborchardt I'm ok with that. There is one change that I think we should still make though and that's making the quote text not bold. So I'll revert my other changes except for that one and merge this if ok with you guys.

codeviking:
@jonborchardt we're actually in agreement.  Let's make this `ai2-web` specific, and we can have a more holistic discussion about how to handle this in Varnish proper.

aaronsarnat:
Merging on green (just the quote font-weight change)

PR #273, opened 2020-02-03
Jon/tags

jonborchardt:
preliminary work to before working on merging for fork

* added tags
* updated spacing values


components/Tag.tsx
```diff
@@ -0,0 +1,103 @@
...
+    &.ant-tag-blue {
+        ${tagStyles.blue};
+    }
+    &.ant-tag-geekblue {
+        ${tagStyles.geekblue};
+    }
+    &.ant-tag-purple {
+        ${tagStyles.purple};
+    }
+`;
```
codeviking: What's the advantage of splitting out `tagStyles`? Could it just be inlined here (less indirection in this case might be easier to follow)?

theme/spacing.ts
```diff
@@ -48,8 +48,10 @@ export const spacing = {
     md: Spacing.fromPixels('md', 16),
     lg: Spacing.fromPixels('lg', 24),
     xl: Spacing.fromPixels('xl', 36),
-    xxl: Spacing.fromPixels('xxl', 48),
-    xxxl: Spacing.fromPixels('xxxl', 64),
+    xxl: Spacing.fromPixels('xxl (depricated, use xl2)', 48),
```
codeviking:
```suggestion
    xxl: Spacing.fromPixels('xxl (deprecated, use xl2)', 48),
```

theme/spacing.ts
```diff
@@ -48,8 +48,10 @@ export const spacing = {
...
+    xl2: Spacing.fromPixels('xl2', 48),
+    xxxl: Spacing.fromPixels('xxxl (depricated, use xl3)', 64),
```
codeviking:
```suggestion
    xxxl: Spacing.fromPixels('xxxl (deprecated, use xl3)', 64),
```

components/Tag.tsx
jonborchardt: ahh... where i copied them from, the style gets reused for styling tags in dropdowns (not copied over yet)

PR #274, opened 2020-02-03
Jon/remove rems

jonborchardt:
we were supporting using rems all over, when we want to default to px for non font items


demo/src/ComponentGroups.tsx
```diff
@@ -172,7 +172,7 @@ export const componentGroups: AppRouteGroup[] = [
         label: 'Data Display',
         routes: [
             {
-                path: `${componentRoot}/tabs`,
+                path: `${componentRoot}/tags`,
```
jonborchardt: oops

Issue #286, opened 2020-03-16
Hovering primary Varnish buttons with a href doesn't look right.


codeviking:
If a `<Button variant="primary" />` has a `href`,  it's hover state looks like:

![image](https://user-images.githubusercontent.com/801451/76785052-f3f35f80-6771-11ea-8e0c-afd8e3cd2867.png)

I suspect this is not intended, as it's hard to see the text.

codeviking:
I ran into this again today. @jonborchardt @schmmd would you two be ok putting this in the on-call queue. It should be a quick fix, and I've ran into it several times. I imagine end users have too.

codeviking:
Here's the fix I did for the CORD-19 explorer:
https://github.com/allenai/santiago/blob/master/ui/src/components/LinkButton.tsx

aimichal:
Actually, this isn't fixed like it should be. <Button variant="primary" href="/">foo<Button> still renders badly.

PR #294, opened 2020-04-20
white anchor text in primary buttons

aimichal:
This makes anchor text white when used in a primary button.

Before:

image

After:

image

This fixes https://github.com/allenai/varnish/issues/286


components/button.tsx
```diff
@@ -168,5 +168,9 @@ export const Button = styled(({ variant, ...props }) => <AntButton {...props} />
                     )};
             }
         }
+
+        a {
+            ${props => (props.type === 'primary' ? `&, &:hover { color: #ffffff; }` : null)}
```
jonborchardt, line 173: we dont want white
we want the varnish color .... parnish.palette.text.contrast (which is white)

aimichal: 👍

aimichal:
> Do we set an `:active` or `:visited` color on links that we need to override too?

When I hover I see:

> image

When I click and hold it looks like that too.

After visiting, and returning to the button page, the button looks like it did before I clicked.

Is that what you were asking?

codeviking:
Yup, thanks for checking!

PR #300, opened 2020-10-06
move less bundling into the lib

jonborchardt:
this moves the logic of bundling the antd less back into the library


demo/src/App.tsx
```diff
@@ -3,7 +3,7 @@ import { BrowserRouter, Route } from 'react-router-dom';
 import { ThemeProvider } from '@allenai/varnish/theme';
 import { ScrollToTopOnPageChange } from '@allenai/varnish/components';
 
-import './App.less';
+import '@allenai/varnish/theme/theme.less';
```
codeviking, line 6: So, isn't this package `@allenai/varnish`? Where is this sourced from? Will this point at `antd` eventually?

theme/theme.less.ts
```diff
@@ -458,15 +462,13 @@ export const less = `
...
     @border-color-split: ${palette.border.default}; // split border inside a component
-    @border-width-base: 1px; // width of the border for a component
 
     // Backgrounds
     @background-color-light: ${color.N2}; // background of header and selected item
     @background-color-base: ${color.N3}; // Default grey background color
 
     // Buttons
     @btn-font-weight: ${typography.font.weight.regular};
-    @btn-border-width: @border-width-base;
```
codeviking, line 469: Are these intentionally being removed? Seems like a side-effect?

demo/src/App.tsx
jonborchardt: no.. this is sourced from theme.less.tx (which points to antd)

theme/theme.less.ts
jonborchardt: no, i added them because they were needed since i WAS removing antd references... but with it back in the lib, these are no longer needed

theme/theme.less.ts
```diff
@@ -31,6 +31,10 @@ export const less = `
     // the fonts we use
     @import (css) url("${getFontImportsURL()}");
 
+    // start with ant design less
```
jonborchardt, line 34: @codeviking see here

demo/src/App.tsx
codeviking: How is `@allenai/varnish` mapped to that? Is this via `npm link`?

Given this isn't sample code, do we need to do that? Could we just use a relative path instead, i.e.:

```
import ../../theme/theme.less
```

I feel like that'd be easier to reason about.

theme/theme.less.ts
codeviking: Gotcha, thanks.

demo/src/App.tsx
codeviking: Ok, per offline, this is due to the use of `npm link`.

codeviking (approved):

demo/src/App.tsx
jonborchardt: it does use npm link....
and we do this all over in the code... im inclined to leave it as is for now

PR #301, opened 2020-10-06
bring back icon utils

jonborchardt:
i removed icon when removing antd from lib, not that it is back, i want to reuse these icon wrappers


components/icon.tsx
```diff
@@ -0,0 +1,31 @@
...
+interface ImgIconProps {
+    src: string;
+    alt?: string;
+    ariaHidden?: boolean;
+}
+export class ImgIcon extends React.PureComponent<ImgIconProps> {
+    render() {
+        const { src, alt = '', ariaHidden, ...other } = this.props;
+        return (
+            <AntdIcon
```
codeviking, line 13: What's the advantage of using `AntdIcon` with an image that we provide?

components/icon.tsx
```diff
@@ -0,0 +1,31 @@
...
+
+interface SvgIconProps {
+    Svg: React.FunctionComponent<React.SVGProps<SVGSVGElement>>;
+}
+export class SvgIcon extends React.PureComponent<SvgIconProps> {
+    render() {
+        const { Svg, ...other } = this.props;
+        return <AntdIcon {...other} component={() => <Svg />} />;
+    }
+}
```
codeviking, line 31: Is this to just support custom icons? Do we need that still, I thought we were just going to use `antd`'s built in icons, since we don't have an icon designer on staff anymore?

components/icon.tsx
```diff
@@ -0,0 +1,31 @@
...
+                component={() => (
+                    <img src={src} alt={alt} aria-hidden={ariaHidden ? 'true' : 'false'} />
+                )}
+            />
+        );
+    }
+}
+
+interface SvgIconProps {
+    Svg: React.FunctionComponent<React.SVGProps<SVGSVGElement>>;
```
codeviking, line 24: I'm not used to seeing properties with an uppercase character. Maybe this is just `svg`?

components/icon.tsx
```diff
@@ -0,0 +1,31 @@
+import * as React from 'react';
+import AntdIcon from '@ant-design/icons';
+
+interface ImgIconProps {
+    src: string;
+    alt?: string;
+    ariaHidden?: boolean;
+}
+export class ImgIcon extends React.PureComponent<ImgIconProps> {
```
aimichal, line 9:
```suggestion
}

export class ImgIcon extends React.PureComponent<ImgIconProps> {
```

codeviking (changes requested):
I don't understand why we need this component. Can't we just use `antd`'s icons without any additional code? Why do we need these wrapper elements, which appear to do very little?

components/icon.tsx
jonborchardt: if we want to ever ad an svg to a button or whatever, we need this
we use this today on prior-demo and allennlp-demo
we can remove it... but its generally useful to be able to add an svg to a button

components/icon.tsx
jonborchardt: the Svg gets used asa component

jonborchardt:
we can discuss removing this later, but it is needed to get parity with current code bases... this code would need to be added prior-demo and allennlp-demo and anywhere someone wants to add a svg to a button (like in beaker likely)

components/icon.tsx
jonborchardt: if we want to ever ad an svg to a button or whatever, we need this

components/icon.tsx
codeviking: Sure, but why do we need this? What do we gain by using `AntDIcon`? That's what I don't understand.

codeviking:
> it is needed to get parity with current code bases... 

I didn't think we were shooting for pure backwards compatibility, since this is a major revision of Varnish. Maybe it's ok to break things, if it makes sense too?

codeviking:
In fact, I think it is the right thing to do.  It appears `antd` clearly documents how to use custom icons:
https://ant.design/components/icon/#components-icon-demo-custom

So, could sites that need to do ☝️  simply use that example code, rather than our custom component? One less thing for us to maintain (particularly since it doesn't really do much other than alias the `component` property to `svg`).

jonborchardt:
> In fact, I think it is the right thing to do. It appears `antd` clearly documents how to use custom icons:
> https://ant.design/components/icon/#components-icon-demo-custom
> 
> So, could sites that need to do ☝️ simply use that example code, rather than our custom component? One less thing for us to maintain (particularly since it doesn't really do much other than alias the `component` property to `svg`).

we can remove this, BUT it will need to be added AS IS here to mosaickg, allennlp-demo, varnish demo, and prior-demo...
so i dont see hte harm in including it here.

jonborchardt:
ok, removing it

codeviking (approved):
LGTM!

PR #302, opened 2020-10-08
add mdx support and show how to use it on ai2logos

jonborchardt:
this is a working example of using mdx to drive the demos for the varnish docs
only ai2logos have been converted
after this pr, ill convert all others

image

image

image


@types/mdx.d.ts
```diff
@@ -0,0 +1,45 @@
...
+    export type Components = {
+        [key in ComponentType]?: React.ComponentType<{ children: React.ReactNode }>;
+    };
+    export interface MDXProviderProps {
+        children: React.ReactNode;
+        components: Components;
+    }
+    export class MDXProvider extends React.Component<MDXProviderProps> {}
+    let mdx: any; // who knows
+    export { mdx };
```
codeviking, line 44: Would this package help?

https://www.npmjs.com/package/@types/mdx-js__react

Maybe it'd allow us to omit this file, or at least provide a type signature for `mdx`?

@types/mdx.d.ts
```diff
@@ -0,0 +1,45 @@
...
+        | 'ol'
+        | 'p'
+        | 'pre'
+        | 'strong'
+        | 'sup'
+        | 'table'
+        | 'td'
+        | 'thematicBreak'
+        | 'tr'
+        | 'ul';
```
codeviking, line 34: Are these MDX provided things? `thematicBreak`, for instance?

demo/src/App.tsx
```diff
@@ -2,10 +2,25 @@ import * as React from 'react';
...
+
+// a mapping of components overrides used by MDX
+const components = {
+    code: (props: any) => <LiveProvider {...props} />,
+    table: (props: any) => (
+        <TableWrapper>
+            <table {...props} />
+        </TableWrapper>
+    ),
+    Examples: Examples,
```
codeviking, line 21: What's this mapping? Do we have an example (hah!) of how it's used?

demo/src/ComponentGroups.tsx
```diff
@@ -23,6 +22,48 @@ import spacingIcon from './icons/spacing-14px.svg';
...
+        cols: number;
+        iconSrc: string;
+    };
+    default: (props: any) => any; // TODO, JSX typing is not importing for me for some reason
+}
+
+// turn titles into urls
+const toLowerReplaceSpace = (title: string) => {
+    return title.replace(' ', '_').toLowerCase();
+};
```
codeviking, line 45: There's other characters that'll trip up URLs too. You might consider using something like [this package](https://www.npmjs.com/package/slug) to do it for you. You could also just wait until we hit those edge-cases and change this function as needed.

That said I might just call this `toURLSlug` or `toURLSafeName` or something like that, since I think that's the intent.

demo/src/ComponentGroups.tsx
```diff
@@ -23,6 +22,48 @@ import spacingIcon from './icons/spacing-14px.svg';
 import breakpointIcon from './icons/breakpoint-14px.svg';
 import headerFooterIcon from './icons/header-footer-14px.svg';
 
+const componentRoot = '/components';
+
+// find all index mdx pages to show in menu
+const mdxs = (require as any).context('./pages', true, /index\.(mdx)$/);
```
codeviking, line 28: Does this load them at build time or runtime?

demo/src/ComponentGroups.tsx
```diff
@@ -23,6 +22,48 @@ import spacingIcon from './icons/spacing-14px.svg';
...
+        label: t,
+        routes: examples
+            .filter((ex) => ex.metadata.type === t)
+            .sort((a, b) => a.metadata.title.localeCompare(b.metadata.title))
+            .map((ex) => {
+                return {
+                    path: `${componentRoot}/${toLowerReplaceSpace(ex.metadata.title)}`,
+                    label: ex.metadata.title,
+                    iconSrc: ex.metadata.iconSrc,
+                    component: ex.default as any, // TODO, JSX typing is not importing for me for some reason
```
codeviking, line 61: I'd expect `React.ComponentType` or `React.Component` to work here -- did you try that?

demo/src/Examples.tsx
```diff
@@ -0,0 +1,113 @@
...
+            &:nth-child(3) {
+                width: 22%;
+                color: ${({ theme }) => theme.color.M7};
+                word-break: break-all;
+            }
+            &:nth-child(4) {
+                width: 16%;
+            }
+        }
+    }
```
codeviking, line 87: Is there a reason this isn't just our standard Varnish table?

demo/src/Examples.tsx
```diff
@@ -0,0 +1,113 @@
...
+    font-weight: ${({ theme }) => theme.typography.font.weight.bold};
+`;
+
+const ExampleWrapper = styled.div`
+    padding: ${({ theme }) => theme.spacing.lg};
+    margin-bottom: ${({ theme }) => theme.spacing.xl2};
+    border: 1px solid ${({ theme }) => theme.palette.border.default};
+    // inline-block allows demos to not break across columns
+    display: inline-block;
+    width: calc(100% - ${({ theme }) => theme.spacing.sm});
```
codeviking, line 112: I wonder if we could tweak `box-sizing` to eliminate the need for this.

Or maybe try `width: 100vw` (the `vw` unit is great).

demo/src/pages/AI2Logos/demo/default.mdx
```diff
@@ -0,0 +1,19 @@
+export const metadata = {
+  order: 0,
+  title: 'AI2 Logo',
+}
```
codeviking, line 4: I don't see `order` being used above for sorting -- is this a leftover, maybe?

demo/src/pages/AI2Logos/demo/sizes.mdx
```diff
@@ -0,0 +1,35 @@
+export const metadata = {
+  order: 1,
+  title: 'Sizes',
+}
+
+This component includes a `size` property.
```
codeviking, line 6: Maybe we should document the different options for `size`? i.e.:

```
Size is one of `"lg"`, `"micro"` or `"default"`.
```

demo/src/pages/AI2Logos/demo/text.mdx
```diff
@@ -0,0 +1,19 @@
+export const metadata = {
+  order: 2,
+  title: 'Removing Organization Name Text',
+}
```
codeviking, line 4: Is this a pretty standard way to add this type of information to `mdx` files? Do they by chance use frontmatter? That's how I'd expect this to be expressed.

It looks like it might require a custom loader, which is odd:
https://mdxjs.com/guides/custom-loader

codeviking (approved):
LGTM; The MDX stuff is great. I had a few questions.

@types/mdx.d.ts
jonborchardt: YES, ty, i didnt see that

@types/mdx.d.ts
jonborchardt: moot, removing

demo/src/App.tsx
jonborchardt: this mapping is used when the MDX parses the doc, if it sees something that is not in scope, it gets it from components
the example use is in the mdx file
(it basically extends mdx scope)

@types/mdx.d.ts
jonborchardt: interesting: https://talk.commonmark.org/t/horizontal-rule-or-thematic-break/912

demo/src/Examples.tsx
jonborchardt: the varnish (html) table does not look good... and we wont be able to use the full table in md
does the shellac table look good?

this style looks like this:
image

demo/src/ComponentGroups.tsx
jonborchardt: now using slug, its great

demo/src/ComponentGroups.tsx
jonborchardt: i think build
you cannot call require.context(variable
you must call with the actual directory

demo/src/ComponentGroups.tsx
jonborchardt: ill work on it... i held off because it may cause a larger change... but ill look into it

jonborchardt: React.ReactElement for the win

demo/src/Examples.tsx
jonborchardt: the calc was for something old.. removed it

demo/src/pages/AI2Logos/demo/default.mdx
jonborchardt: examples are sorted on order
`.sort((a, b) => a.metadata.order - b.metadata.order)`

demo/src/pages/AI2Logos/demo/sizes.mdx
jonborchardt: image

demo/src/pages/AI2Logos/demo/text.mdx
jonborchardt: ill look into it...
the mdx files can import and export anything they want...
moving to frontmater does make the md cleaner

@types/mdx.d.ts
jonborchardt: fix pushed

@types/mdx.d.ts
jonborchardt: removed

demo/src/ComponentGroups.tsx
jonborchardt: fix pushed

demo/src/Examples.tsx
jonborchardt: fix pushed

demo/src/pages/AI2Logos/demo/text.mdx
jonborchardt: updated

PR #303, opened 2020-10-10
add breakpoint mdx

jonborchardt:


demo/src/App.tsx
```diff
@@ -18,8 +19,9 @@ const components = {
             <table {...props} />
         </TableWrapper>
     ),
-    Examples: Examples,
+    Examples,
     hr: Divider,
+    BreakpointDisplay,
```
jonborchardt, line 24: any react component you want to show in mdx needs to be added here

demo/src/Examples.tsx
```diff
@@ -24,14 +24,14 @@ export const Examples = (props: { examples: any[]; cols: number }) => (
...
-export const LiveProvider = (props: { children: string; live: any }) => {
+export const LiveProvider = (props: { children: string; live: boolean }) => {
     if (props.live) {
         return (
             <DefaultLiveProvider
                 code={props.children.trim()}
                 transformCode={(code: string) => '/** @jsx mdx */' + code.split('//--')[1]}
                 scope={{ mdx }}
-                noInline={false}
+                noInline={true}
```
jonborchardt, line 34: now doing noinline for all demos for simplicity

demo/src/pages/AI2Logos/demo/default.mdx
```diff
@@ -11,7 +11,7 @@ Basic usage.
 import { AI2Logo } from '@allenai/varnish/components';
 //--
 
-() => (
+render(
```
jonborchardt, line 14: this is noinline

demo/src/shared/DefaultLiveProvider.tsx
```diff
@@ -28,6 +29,7 @@ const globalScope = {
     ...components,
     icon: icon,
     Menu,
+    belowOrEqualTo,
```
jonborchardt, line 32: if you want the live code editor to have an import, you must add it here

codeviking:
Should this wait for review prior to merging?

demo/src/App.tsx
codeviking: Bummer. Maybe we should have a document for "how to add a new thing to Varnish..."?

demo/src/App.tsx
```diff
@@ -9,6 +9,7 @@ import '@allenai/varnish/theme/theme.less';
 import AppChrome from './AppChrome';
 import { Divider } from './shared';
 import { LiveProvider, TableWrapper, Examples } from './Examples';
+import { BreakpointDisplay } from './pages/Breakpoints/breakpointDisplay';
```
codeviking, line 12: What's `BreakpointDisplay` mean? What's the `Display` suffix for?

demo/src/Examples.tsx
codeviking: What does this mean?

demo/src/pages/AI2Logos/demo/default.mdx
codeviking: So, we have to declare a `render()` function? Why the change?

demo/src/pages/Breakpoints/breakpointDisplay.tsx
```diff
@@ -0,0 +1,59 @@
+import * as React from 'react';
+import styled from 'styled-components';
+import { Spacing } from '@allenai/varnish/theme/spacing';
+import { breakpoints } from '@allenai/varnish/theme/breakpoints';
+
+export class BreakpointDisplay extends React.PureComponent {
+    render() {
+        return (
+            <BreakpointGrid>
+                {Object.keys(breakpoints).map((bKey) => {
```
codeviking, line 10: Consider using `Object.entries()` here instead:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/entries

```
Object.entries(breakpoints).map(([ breakpointKey, breakpoint ]) => {
   // ...
});
```

demo/src/pages/Breakpoints/demo/default.mdx
```diff
@@ -0,0 +1,61 @@
+---
+order: 0
+title: Basic Usage
+---
+
+Change the screen width to see the color change based on breakpoints.
```
codeviking, line 6: We should probably explain what a breakpoint is, since our users probably aren't familiar.

Maybe something like:

> Breakpoints are a series of pixel values which represent different screen widths at which behavior can be changed to accommodate different sized screens.  For instance, below a certain breakpoint you might change the layout so that it's single column instead of two column so that the page is easier to use on a cellphone.

demo/src/pages/Breakpoints/demo/default.mdx
```diff
@@ -0,0 +1,61 @@
...
+
+```jsx live
+import styled from 'styled-components';
+import { belowOrEqualTo } from '@allenai/varnish/theme/breakpoints';
+//--
+
+const BreakpointDiv = styled.div`
+    width: 180px;
```
codeviking, line 14: Should we be using `width`, since normally we want to discourage users from setting this? "Content is King"...

demo/src/pages/Breakpoints/demo/default.mdx
```diff
@@ -0,0 +1,61 @@
...
+order: 0
+title: Basic Usage
+---
+
+Change the screen width to see the color change based on breakpoints.
+
+```jsx live
+import styled from 'styled-components';
+import { belowOrEqualTo } from '@allenai/varnish/theme/breakpoints';
+//--
```
codeviking, line 11: What's this?

demo/src/pages/Breakpoints/demo/default.mdx
```diff
@@ -0,0 +1,61 @@
...
+const BreakpointDiv = styled.div`
+    width: 180px;
+    height: ${({theme}) => theme.spacing.lg};
+    margin: ${({theme}) => theme.spacing.md};
+    padding: 0 ${({theme}) => theme.spacing.sm};
+    background: ${({theme}) => theme.color.N6};
+
+    ::after {
+      content: '> breakpoints.xl2'
+    }
```
codeviking, line 22: I'd propose we omit pseudo selectors and just write out the HTML manually. I think pseudo selectors are a bit of an anti-pattern, in that they're "side-effects" of the document styles.

demo/src/pages/Breakpoints/index.mdx
```diff
@@ -0,0 +1,25 @@
+---
```
codeviking, line 1: What's the different between this file and the `/default.mdx` one? Why do we have both?

demo/src/pages/Breakpoints/index.mdx
```diff
@@ -0,0 +1,25 @@
...
+cols: 1
+iconSrc: ./breakpoint-14px.svg
+---
+import defaultDemo from './demo/default.mdx'; // crazy, we must import for the require to work
+const mdxs = require.context('./demo', false, /\.(mdx)$/);
+const examples = mdxs.keys().map(mdxs);
+
+## Breakpoints
+
+Standard breakpoints values.
```
codeviking, line 13: What does `Standard` mean? I think these were ones we selected with Nick's guidance, but I'm not sure if they align with any standard or best practices, right?

codeviking (commented):
I think we should expand the documentation a bit, and proposed a few amendments.

demo/src/App.tsx
jonborchardt: will do, also i made this cleaner, now auto including all from /pages/..index.ts

demo/src/Examples.tsx
jonborchardt: it means we must have a render method, AND we can do stuff before the render...
false assumes wee are INSIDE a render

demo/src/pages/AI2Logos/demo/default.mdx
jonborchardt: see above

demo/src/App.tsx
jonborchardt: changed to values

demo/src/pages/Breakpoints/demo/default.mdx
jonborchardt: ?

demo/src/pages/Breakpoints/demo/default.mdx
jonborchardt: its a hack, ill document it

the stuff above the //-- is not parsed into the react live view

demo/src/pages/Breakpoints/index.mdx
jonborchardt: index holds the doc for X
default holds the most basic demo inside of index
(these names are to be consistent with antd so i can bring over more antd if need be)

demo/src/App.tsx
jonborchardt: fixed in color pr

demo/src/App.tsx
jonborchardt: fixed in color pr

demo/src/pages/Breakpoints/breakpointDisplay.tsx
```diff
@@ -0,0 +1,59 @@
+import * as React from 'react';
```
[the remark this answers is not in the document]
jonborchardt: fixed in color pr

demo/src/pages/Breakpoints/breakpointDisplay.tsx
jonborchardt: fixed in color pr

demo/src/pages/Breakpoints/demo/default.mdx
jonborchardt: fixed in color pr

demo/src/pages/Breakpoints/demo/default.mdx
jonborchardt: fixed in color pr

demo/src/pages/Breakpoints/index.mdx
jonborchardt: fixed in color pr

demo/src/pages/Breakpoints/demo/default.mdx
jonborchardt: this is the cleanest way to show whats going on in a demo... do you have another proposal?

demo/src/Examples.tsx
codeviking: Gotcha. Makes sense, thanks.

demo/src/pages/AI2Logos/demo/default.mdx
codeviking: Thanks. Yup!

demo/src/pages/Breakpoints/demo/default.mdx
codeviking: I think we discourage people from hard-coding the dimensions of things, so that it's sized based upon it's content, right?  Doing this helps avoid weird bugs where as content-changes the surrounding layout doesn't.

So I'm asking if we should omit using `width` in our examples, as to practice what we preach -- maybe it's unavoidable here, I'm not sure.

demo/src/pages/Breakpoints/demo/default.mdx
codeviking: Gotcha, thanks.

demo/src/pages/Breakpoints/demo/default.mdx
codeviking: I'd just add the content to the JSX itself, something like:

```jsx
<SomeBreakpointThing /><Label> > breakpoints.xl2 </Label>
```

demo/src/pages/Breakpoints/index.mdx
codeviking: Gotcha. Thanks.

demo/src/pages/Breakpoints/demo/default.mdx
jonborchardt: the point of this demo is to show that the breakpoint can affect the css...
im really not sure how your example would work here

demo/src/pages/Breakpoints/index.mdx
jonborchardt: k

demo/src/pages/Breakpoints/demo/default.mdx
jonborchardt: gotcha
making 100% (there is no content)

demo/src/pages/Breakpoints/demo/default.mdx
codeviking: Sure, ok. Let's not worry about it, we can revisit later if this comes up.

PR #304, opened 2020-10-11
add color mdx demos

jonborchardt:
add mdx for color


demo/src/pages/Colors/colorDisplay.tsx
```diff
@@ -1,12 +1,12 @@
...
-import { RouteComponentProps } from 'react-router';
+import { Typography } from 'antd';
 import { BodyJumbo, Body, BodyMicro } from '@allenai/varnish/components';
 import { Theme, Color } from '@allenai/varnish/theme';
 import { dictionaryToArray } from '@allenai/varnish/utils/base';
 
-import { PageTitle, DefaultLiveProvider, SectionWithDivider } from '../shared';
-const { color, chartingColor } = Theme.default;
+const DefaultTheme = Theme.default;
+const { color, chartingColor } = DefaultTheme;
```
codeviking, line 9: I think we should consider improving the accessibility of our theme. We don't export any other themes beyond the `default`, and I'm not sure if we plan to.  I'd propose we do something like:

```
import { Theme } from `@allenai/varnish`
```

...or if we do plan on having other themes:

```
import { theme } from '@allenai/varnish'

// Usage: theme.Default
```

demo/src/pages/Colors/demo/contrast.mdx
```diff
@@ -0,0 +1,30 @@
...
+`;
+
+const colorsToRender = ['B1','B2','B3','B4','B5','B6','B7','B8','B9','B10'];
+
+render(
+  <div>
+    {colorsToRender.map(color => <ColorDiv key={color} col={color}>{color}</ColorDiv> )}
+  </div>
+);
+```
```
codeviking, line 30: Why do we need the `useContractText` property, and how does it get toggled? Is there some magic that changes it based upon the background color?

demo/src/pages/Colors/demo/indirect.mdx
```diff
@@ -0,0 +1,26 @@
+---
```
codeviking, line 1: What's an indirect color? Maybe this is just called `palette.mdx`, or we just document this alongside other documentation for the `palette`?

demo/src/pages/Colors/demo/indirect.mdx
```diff
@@ -0,0 +1,26 @@
...
+    background: ${(props) => props.theme.palette.background.error};
+    border: ${(props) => `1px solid ${props.theme.palette.border.error}`};
+`;
+
+render(
+  <div>
+    <Error>I cant do that Dave.</Error>
+  </div>
+);
+```
```
codeviking, line 26: It's a little confusing right now as to when one should use `palette` or `color`. I find myself just using `color` and rarely using `palette`, since the `color` API has a predictable structure while the `palette` is a series of named keys that are hard to remember.

Is `palette` important to keep? Maybe we drop it for simplicity? Or, we could merge the documentation into one and explain when to use one versus the other.  Either way I think we should answer this question, as I myself am confused right now, so I bet our users are too :).

codeviking (approved):
LGTM. I think we should try to answer the question of when one should use `palette` vs `color`.

I find the `palette` abstraction harder to use in my experience. I tend to just use `color`. Maybe I shouldn't be doing that, but maybe the whole `palette` thing isn't what we thought it would be (in which case maybe we drop it -- less for us to maintain, and less for our users to think about).

demo/src/pages/Colors/demo/contrast.mdx
jonborchardt: it is defined per color as part of the theme...
it is needed because you need to be able to know if varnish suggests using the contrast text color on this color or not

demo/src/App.tsx
```diff
@@ -10,6 +10,7 @@ import AppChrome from './AppChrome';
 import { Divider } from './shared';
 import { LiveProvider, TableWrapper, Examples } from './Examples';
 import { BreakpointDisplay } from './pages/Breakpoints/breakpointDisplay';
+import { PrimaryColorDisplay, ExtendedColorDisplay } from './pages/Colors/colorDisplay';
```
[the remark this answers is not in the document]
jonborchardt: display is just what antd called them... i agree, and il rename

demo/src/pages/Colors/demo/indirect.mdx
jonborchardt: we need to keep it, as we dont want to have N7 all over the code for all borders...
if we decide to change borders to N8, we need to understand the N7 usage in many spots, where, theme.border is understandable...

ill try to clean up the docs to make this clear

jonborchardt: i will do this work when i do palette later today

demo/src/App.tsx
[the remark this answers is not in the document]
jonborchardt: fixed

demo/src/pages/Colors/colorDisplay.tsx
jonborchardt: fixed

demo/src/pages/Colors/demo/indirect.mdx
jonborchardt: fixed

demo/src/App.tsx
```diff
@@ -9,10 +9,20 @@ import '@allenai/varnish/theme/theme.less';
 import AppChrome from './AppChrome';
 import { Divider } from './shared';
 import { LiveProvider, TableWrapper, Examples } from './Examples';
-import { BreakpointDisplay } from './pages/Breakpoints/breakpointDisplay';
+
+// find all index ts components to add to mdx context
```
jonborchardt, line 13: this allows us to not have to specify the new components as part of the mdx context

demo/src/App.tsx
```diff
@@ -9,10 +9,20 @@ import '@allenai/varnish/theme/theme.less';
...
+const valueTs = (require as any).context('./pages', true, /index\.(ts)$/);
+const valueComponents = valueTs.keys().map(valueTs);
+const valueJson: any = {};
+valueComponents.forEach((module: any) => {
+    Object.keys(module).forEach((k: string) => {
+        valueJson[k] = module[k];
+    });
+});
```
codeviking, line 21: This was a little hard to follow. I looked into the `context` function and I think I understand. I might break this out into a function with some additional documentation to make it easier to follow:

```TypeScript
// Dynamically imports all symbols from files with the name `index` in the `pages` directory.
// We use this so that the MDX components can use Components that's defined there.
// TODO: @jonb I'm not sure on the types here.
function loadPageComponents(): { [symbolName: string]: {} } {
    const moduleLoaderByPath: { [path: string]: () => {} } = (request as any).context('./pages', true, /index\.(ts)$/)
    const importedSymbols = { [symbolName: string]: {} };
    for [ path, loadModule ] of Object.entries(moduleLoaderByPath) {
        const exportedSymbols = loadModule(path);
        for [ name, value ] of Object.entries(exportedSymbols) {
             if (importedSymbols has name) { 
                   throw Error(`The ${name} symbol defined in ${path} is exported by something else.`);
             }
             importedSymbols[name] = value;
        }
    } 
    return importedSymbols;
}
```

Then you'd use it like `...loadPageComponents()`.

Also, I wonder if we could use [this package](https://www.npmjs.com/package/@types/webpack) so that `require` and the `context` function is typed? Maybe we are, and the `any` cast is still necessary?

demo/src/Examples.tsx
```diff
@@ -29,7 +29,10 @@ export const LiveProvider = (props: { children: string; live: boolean }) => {
         return (
             <DefaultLiveProvider
                 code={props.children.trim()}
-                transformCode={(code: string) => '/** @jsx mdx */' + code.split('//--')[1]}
+                transformCode={(code: string) =>
+                    // remove import statements
+                    '/** @jsx mdx */' + code.replace(/^import.*$/gm, '')
```
codeviking, line 34: I actually like the original better, this would remove variables or functions that start with `import`, i.e. if I wrote a function called `importStyles`...

The prior one is less error prone, though I might make it more explicit:

You could declare that the introductory content is called a `premable`, and indicate it's conclusion:
```
// -- end preamble --
```

Or you could simply declare where the example starts:

```
// -- begin example ---
```

demo/src/pages/Colors/demo/contrast.mdx
codeviking: Ok, so I think I follow now.

The `useContrastText` is attached to a `Color` instance, and can be used to dictate whether the text should be light or dark.

Is that right?

demo/src/pages/Colors/demo/contrast.mdx
```diff
@@ -0,0 +1,35 @@
+---
+order: 2
+title: Text Contrast
+---
+
+When adding text on top of a color, we need to know if we should use text that is the default color
+(dark grey) or the contrast color (off white).  This ensures proper contrast and readability.
+
+To properly color text on top of a Varnish color, use the `useContrastText` property to pick the
+correct text color.
```
codeviking, line 10: If my explanation below is accurate, maybe we just say:

> Each color defined by Varnish has a `useContrastText` boolean attached to it. If `true`, and the color is used as the `background` of an element, we suggest setting the text color to `palette.text.contrast` for readability.

codeviking (approved):
LGTM. I proposed a few small tweaks.

demo/src/pages/Colors/demo/contrast.mdx
jonborchardt: yes
see demo

demo/src/Examples.tsx
jonborchardt: what if i look for `import ` <- with a space

codeviking: Yea, that should work.

I favor the `//--` syntax because it's very explicit, and very safe. Any line that beings with `//` is a comment, so removing it should have no effect on the code :).

demo/src/App.tsx
jonborchardt: i could not get htis to work as is, but i did make a function and better comments

demo/src/pages/Colors/demo/contrast.mdx
jonborchardt: sounds good

PR #306, opened 2020-10-13
add spacing mdx

jonborchardt:


demo/src/pages/Breakpoints/demo/default.mdx
```diff
@@ -10,7 +10,7 @@ import styled from 'styled-components';
 import { belowOrEqualTo } from '@allenai/varnish/theme/breakpoints';
 
 const BreakpointDiv = styled.div`
-    width: 100%;
+    width: 90%;
```
codeviking, line 13: Is this actually `%` (in that it's of it's parent), or is this a place where we might want to use `vw` units?

demo/src/pages/Spacing/SpacingValues.tsx
```diff
@@ -0,0 +1,84 @@
...
+        return (
+            <>
+                <Namespace>
+                    <BodyMicro>theme.spacing.</BodyMicro>
+                </Namespace>
+                <SpacingGrid>
+                    {Object.keys(spacing).map((sKey) => {
+                        const spac = spacing[sKey];
+                        return <SpacingRow spacing={spac} key={sKey} label={sKey} />;
+                    })}
```
codeviking, line 18: I think `Object.entires` could work here too.

demo/src/pages/Breakpoints/demo/default.mdx
jonborchardt: yes of parent....

demo/src/pages/Spacing/SpacingValues.tsx
jonborchardt: will do

demo/src/pages/Spacing/page.mdx
```diff
@@ -0,0 +1,26 @@
...
+iconSrc: ./spacing-14px.svg
+---
+
+import defaultDemo from './demo/default.mdx'; // crazy, we must import for the require to work
+const mdxs = require.context('./demo', false, /\.(mdx)$/);
+const examples = mdxs.keys().map(mdxs);
+
+## Spacing
+
+Standard AI2 spacing units.
```
codeviking, line 14:
```suggestion
Standard AI2 spacing units. Use these values to create space between and/or inside of elements. If you're setting `margin` or `padding`, you should probably be using one of these values. We've selected values that when used together create a consistent visual rhythm, making the resulting site look better and easier to scan.
```

codeviking (approved):

PR #307, opened 2020-10-14
added link mdx

jonborchardt:
this adds link to mdx demos, deletes some unused code, and updates footer and context to get link to work


components/Footer.tsx
```diff
@@ -51,27 +43,18 @@ export class Footer extends React.PureComponent<Props> {
...
-                                </ExternalLink>
-                            </Body>
+                                <a href="https://allenai.org/privacy-policy.html">Privacy Policy</a>{' '}
+                                | <a href="https://allenai.org/terms.html">Terms of Use</a>
+                            </span>
                         )}
                     </StyledFooter>
                 )}
-            </LayoutContext.Consumer>
+            </AppLayoutContext.Consumer>
```
codeviking, line 57: We're probably dropping `AppLayoutContext`, right?

components/Link.tsx
```diff
@@ -0,0 +1,20 @@
...
+
+export function contrastLinkColorStyles() {
+    return css<{ active?: boolean }>`
+        color: ${({ theme, active }) =>
+            active ? theme.link.contrastActiveColor : theme.link.contrastColor};
+        &:hover {
+            color: ${({ theme }) => theme.link.hover.contrastColor};
+        }
+    `;
+}
```
codeviking, line 20: Hmm. This is odd. We're emitting css-only? How do we expect users to use this?

demo/src/App.tsx
```diff
@@ -30,6 +30,7 @@ const loadPageComponents = (): { [symbolName: string]: {} } => {
 // a mapping of components overrides used by MDX
 const components = {
     ...loadPageComponents(),
+    pre: (props: any) => <span {...props} />,
```
codeviking, line 33: Why doesn't a `<pre />` become a `<pre />` tag?

demo/src/AppChrome.tsx
```diff
@@ -198,3 +198,7 @@ const IconMenuItem = ({ route, menuCollapsed }: { route: AppRoute; menuCollapsed
         </InternalLink>
     );
 };
+
+const InternalLink = styled(Link)`
+    ${linkColorStyles()};
+`;
```
codeviking, line 204: Interesting. Will users have to do this anytime they use a react-router Link? Do we plan to do this for them as part of the `@allenai/varnish-react-router` package?

demo/src/pages/Links/demo/contrast.mdx
```diff
@@ -0,0 +1,28 @@
...
+  color: #fff;
+`;
+
+render(
+  <Contrast>
+    Lorem ipsum dolor sit amet, consectetur adipiscing elit. Vivamus a molestic metus.{' '}
+    <a href="./">Contrast Link</a> lobortis varius. Cras vulputate felis et mauris tincidunt,
+    elementum volutpat.
+  </Contrast>
+);
```
codeviking, line 28: Something feels off here. I'd expect the `<Constrast />` component to set the `color` for me, in addition to the `Link` styles.

Is this something we anticipate users doing often? If so, maybe the `<Contrast />` component is something we should offer. Maybe it's `Typography.Contrast` or something like that, to conform with the `Typopgraphy` convention put forward by `antd`?

Put differently the majority of this documentation is code we're asking them to write, which seems odd.

I also wonder if we're putting the cart before the horse by even implementing the `contrastLinkColorStyles` bit. Do we have a clear use for this right now?

components/Link.tsx
codeviking: Ok. I follow now. It still doesn't sit quite right with me.   

First off, the file probably shouldn't be called `Link`, as there's no `Link` component that's exported. I also think the names are a little duplicative. What if we did something like this:

```
components/LinkStyle.tsx

export LinkStyle {
   default() {
       // Return default link styles
   }
   contrast() {
       // Return contrast link styles
   }
}
```

Usage might look like:

```
import { LinkStyle } from '@allenai/varnish';

const MyLink = styled.a`
   ${LinkStyle.default()}
`;
```

demo/src/pages/Links/page.mdx
```diff
@@ -0,0 +1,46 @@
+---
+type: General
+title: Link
+cols: 2
+iconSrc: ./links-14px.svg
+---
+import defaultDemo from './demo/default.mdx'; // crazy, we must import for the require to work
+const mdxs = require.context('./demo', false, /\.(mdx)$/);
+const examples = mdxs.keys().map(mdxs);
```
codeviking, line 9: What's this?

demo/src/pages/Links/page.mdx
```diff
@@ -0,0 +1,46 @@
...
+
+## Links
+
+Varnish has default styles for `a[href]` links and provides a base style `contrastLinkColorStyles`
+for use on dark backgrounds.
```
codeviking, line 14:
```suggestion
Varnish makes links that use an `<a />` tag look nice by default. These links look nice on light backgrounds. But there's also [a way to make links look nice on dark backgrounds](./link-to-it-here).
```

demo/src/pages/Links/page.mdx
```diff
@@ -0,0 +1,46 @@
...
+
+### Examples
+
+<Examples cols={frontMatter.cols} examples={examples} />
+
+### react-router
+
+`Varnish` does not depend on `react-router`.  If you wish to use `react-router` in your app, you can
+include [`@allenai/varnish-react-router`](https://www.npmjs.com/package/@allenai/varnish-react-router)
+for access to styled router links and other `react-router` dependent components.
```
codeviking, line 24:
```suggestion
If your application uses `react-router` to handle changing pages on the client, you'll want to install `@allenai/varnish-react-router` and use the APIs provided via that package.
```

layout/LayoutContext.tsx
```diff
@@ -9,7 +9,7 @@ import React from 'react';
  *  - The "default" variant is a standard, horizontally centered application. This
  *    is a fairly conventional layout used for content  rich websites.
  */
-export type LayoutVariant = 'app' | 'hcenter';
+export type LayoutVariant = 'app' | 'hcenter'; // TODO: rename AppLayoutVariant
```
codeviking, line 12: Or drop it, in favor of antd's layout mechanism, right?

codeviking (changes requested):
Let's discuss this one a bit. I wonder if:

- We could just use a a few CSS rules.
- The whole contrast based API is necessary -- do we use this anywhere other than the AI2 website footer?

components/Footer.tsx
jonborchardt: unknown yet, it on my list, 
i didnt want to muddy this pr any more than it is
we need appconterxt for the header height change i think....

components/Link.tsx
jonborchardt: sounds good

demo/src/App.tsx
jonborchardt: this has to do with how the live preview works...
i may need to rework it a bit, but ` ` ` adds a pre that messes everything up...
see the middle of this page: https://mdxjs.com/guides/live-code

demo/src/AppChrome.tsx
jonborchardt: i plan on rereleaseing the varnish-react-router as part of the conversion, so this will go away, yes
again, trying to make the pr smaller

demo/src/pages/Links/demo/contrast.mdx
jonborchardt: contrast is just a junk example...
we have no idea where they will use the link... and if they want contrast, they will need to set the style as i have it....
'contrast' is an example of their random container that is dark... ill rename to make more clear... there is no Varnish Contrast container.

demo/src/pages/Links/page.mdx
jonborchardt: examples is an array of demos populated my the mdx files in the demo dir

jonborchardt:
> Let's discuss this one a bit. I wonder if:
> 
> * We could just use a a few CSS rules.
> * The whole contrast based API is necessary -- do we use this anywhere other than the AI2 website footer?

this is needed as we have prescribed how a link looks on a dark background...
varnsih currently does not just supply css classes for this kind of thing

demo/src/Examples.tsx
```diff
@@ -37,6 +37,8 @@ export const LiveProvider = (props: { children: string; live: boolean }) => {
                 noInline={true}
             />
         );
+    } else if (props.staticDark) {
+        return <Code variant="dark">{props.children}</Code>;
     }
```
codeviking, line 42: It feels like `static` and `theme` should be different properties:

```
<LiveProvider static={true} theme="dark" />
```

WDYT?

components/Footer.tsx
codeviking: > we need appconterxt for the header height change i think....

Let's dig in when we get there. I bet we can find a way to omit it -- I think that'd be ideal so people aren't confused by having two ways to do the same thing.

demo/src/AppChrome.tsx
codeviking: Gotcha, thanks.

demo/src/pages/Links/demo/contrast.mdx
codeviking: Gotcha, thanks. The revised name makes more sense to me!

demo/src/pages/Links/page.mdx
codeviking: Gotcha, I'm just now following why we have to do this? Do we list all demos (even those not related to Links) in the Links demo?

A comment might be warranted to explain what this is doing, since I haven't seen us do this before in other documentation files.

codeviking:
> varnsih currently does not just supply css classes for this kind of thing

Gotcha, I don't think we export anything that's just `css()` yet either though, right?

I'm wondering if it would make sense to export a `Link` component, to make this work more like the other things we provide. It would be very simple:

```tsk
// TODO: I'm not sure if the type is right, but there should be a type that captures the default properties of a <a /> tag.

interface Props extends HTMLAnchorElementProps {
   contrast?: boolean;
}

export default function Link(props: Props) {
   return <
}

const A = styled.a`
    ...
`;
```

Then `varnish-react-router` could figure out how to extend / re-use this for the `Link` it exports...

How does that sound?

components/Footer.tsx
jonborchardt: that makes sense, but lets push to different pr

demo/src/App.tsx
jonborchardt: ok, to be clear, its only in mdx files
ill add ac omment

demo/src/pages/Links/page.mdx
jonborchardt: ill add a comment, but this is a relative directory... looking for demos in the local demo dir

codeviking: That makes sense. A comment will demystify this :).

demo/src/App.tsx
codeviking: Yup! I understood that (but it's also worth noting in the comment you add!). Thanks!

components/Footer.tsx
codeviking: Sounds good.

demo/src/Examples.tsx
jonborchardt: static is opposite of live...
ill omit the word static

codeviking (approved):
We chatted and the `css()` route makes sense.

PR #309, opened 2020-10-14
move and rename logo imports to match newer varnish

jonborchardt:
the logo code was updated in varnish ant, so back-porting those updates


components/logos/index.ts
```diff
@@ -0,0 +1,7 @@
+import * as AI2Logo from './AI2Logo';
+import * as AI2Projects from './AI2Projects';
+
+export const Logos = {
```
codeviking, line 4: I usually `lowercase` these "namespaces", so that usage looks like:

```
<logos.AllenNLP />`
```

This way there's no confusion about whether `<Logos />` is a component too.

That said I don't feel strongly, and I think this follows `antd`'s precedent, where `Typography` is a similar construct.

codeviking (approved):

components/logos/index.ts
jonborchardt: ya, im happy to move in the right direction... going to varnish v2 will be very breaking

PR #310, opened 2020-10-15
add header mdx

jonborchardt:


components/Header.tsx
```diff
@@ -1,11 +1,10 @@
 import * as React from 'react';
 import { findDOMNode } from 'react-dom';
 import styled from 'styled-components';
+import { Layout } from 'antd';
 
 import { Columns } from './Columns';
-import { LayoutHeader, Content } from './Layout';
-
-import { LayoutContext } from '../layout';
+import { LayoutContext as AppLayoutContext } from '../layout';
```
codeviking, line 7: Do we intend to rename this context to something more generic? Is that coming later?

We could rename it locally (via `as`, *as* you're already doing -- see what I did there!?) to try it out:

```
// This is just an idea. I'm not sure what name should be used.
import { LayoutContext as ViewportContext } from '../layout';
```

components/Header.tsx
```diff
@@ -109,22 +112,37 @@ const Sticky = styled(LayoutHeader)`
...
-    align-items: center;
-`;
-
 export const HeaderTitle = styled.h5`
     margin: 0;
     font-size: 1.9rem;
     line-height: 1.65rem;
     text-transform: none;
-
+    display: inline;
```
codeviking, line 124: Should it be `inline-block`, so that it can have `margin` and/or `padding`?

Also, if it's `inline` I don't think we need to zero out the `margin` (I could be wrong though).

components/Header.tsx
```diff
@@ -109,22 +112,37 @@ const Sticky = styled(LayoutHeader)`
...
     @media (max-width: ${({ theme }) => theme.breakpoints.xs}) {
         font-size: 1.75rem;
         line-height: 1.5rem;
     }
 `;
+
+export const HeaderColumns = styled(Columns)`
+    width: 100%;
+    align-items: center;
+    padding: 11.5px 0;
```
codeviking, line 134: Should we use a `spacing` value?

components/Header.tsx
```diff
@@ -109,22 +112,37 @@ const Sticky = styled(LayoutHeader)`
...
-
+    display: inline;
     @media (max-width: ${({ theme }) => theme.breakpoints.xs}) {
         font-size: 1.75rem;
         line-height: 1.5rem;
     }
 `;
+
+export const HeaderColumns = styled(Columns)`
+    width: 100%;
```
codeviking, line 132: Do we need this? Shouldn't `display: block` suffice?

components/Header.tsx
```diff
@@ -109,22 +112,37 @@ const Sticky = styled(LayoutHeader)`
...
+    align-items: center;
+    padding: 11.5px 0;
+    a[href] {
+        img {
+            display: inline;
```
codeviking, line 137: I think we want `inline-block` too. Otherwise I don't think the entire image is clickable (to trigger the link).

components/Header.tsx
```diff
@@ -109,22 +112,37 @@ const Sticky = styled(LayoutHeader)`
...
         line-height: 1.5rem;
     }
 `;
+
+export const HeaderColumns = styled(Columns)`
+    width: 100%;
+    align-items: center;
+    padding: 11.5px 0;
+    a[href] {
+        img {
```
codeviking, line 136: The logo could be SVG, or even an emoji 😉.  

I think we may want to break this out a bit so it's more flexible.  I also think it might be nice to just add these as properties of `Header`, rather than duplicating the name there. Usage might look like this:

```tsx
<Header>
  <Header.Columns>
    <Header.Logo href="/">
       <svg>...</svg>
     </Header.Logo>
     <Header.Menu>...</Header.Menu>
  </Header.Columns>
</Header>
```

The `<Header.Logo />` would render an `<a />` tag if there's a `href`, and omit it if not.

WDYT?

components/Header.tsx
```diff
@@ -109,22 +112,37 @@ const Sticky = styled(LayoutHeader)`
...
+            display: inline;
+            margin-right: ${({ theme }) => theme.spacing.xs};
+        }
+        ${HeaderTitle} {
```
codeviking, line 140: This could be `<Header.Title />` too.

demo/src/pages/Header/demo/default.mdx
```diff
@@ -0,0 +1,20 @@
+---
+order: 0
+title: Default
+---
+
+Simplest Usage.
```
codeviking, line 6: Maybe we use `Default` instead of `Simplest`? It still doesn't sit right with me for some reason.  😅

demo/src/pages/Header/demo/default.mdx
```diff
@@ -0,0 +1,20 @@
...
+
+```jsx live
+import { Header, HeaderColumns, HeaderTitle } from '@allenai/varnish/components';
+
+render(
+  <Header>
+    <HeaderColumns gridTemplateColumns="auto 1fr">
```
codeviking, line 13: What about simplifying `gridTemplateColumns` to just `columns`. I don't think the user is actually concerned with the fact that it uses `display: grid` and that this is the `grid-template` property. Those are implementation details.

So we'd do:

```
<Header.Columns columns="auto 1fr" />
```

Also `antd` has a [Grid](https://ant.design/components/grid/) component, should we be using that instead?

demo/src/pages/Header/demo/justBanner.mdx
```diff
@@ -0,0 +1,14 @@
...
+
+Just using the AI2 Banner.
+
+```jsx live
+import { AI2Banner } from '@allenai/varnish/components';
+
+render(
+  <AI2Banner />
+);
+```
```
codeviking, line 14: 🎉  Nice. I'm glad this is possible now!

demo/src/pages/Header/demo/justBanner.mdx
```diff
@@ -0,0 +1,14 @@
+---
+order: 3
+title: AI2 Banner
+---
+
+Just using the AI2 Banner.
```
codeviking, line 6:
```suggestion
This renders an AI2 banner on the top of your site. It's required for all AI2 sites.  Link back to the mothership, young skywalker.
```

demo/src/pages/Header/demo/withLogo.mdx
```diff
@@ -0,0 +1,25 @@
+---
+order: 1
+title: With Logo and Content
+---
+
+Header inside a layout with a logo and content.
```
codeviking, line 6:
```suggestion
This shows how to render a header with an application logo. Logos make your application more fun, even emojis suffice!
```

demo/src/pages/Header/demo/withMenu.mdx
```diff
@@ -0,0 +1,44 @@
+---
+order: 1
+title: With Menu
+---
+
+Header inside a layout with a logo, menu, and content.
```
codeviking, line 6:
```suggestion
This shows how to render a header with a logo and navigation.
```

demo/src/pages/Header/page.mdx
```diff
@@ -0,0 +1,28 @@
...
+iconSrc: ./header-footer-14px.svg
+---
+import defaultDemo from './demo/default.mdx'; // crazy, we must import for the require to work
+// these two lines grab the mdx files in the local demo dir and make examples for each
+const mdxs = require.context('./demo', false, /\.(mdx)$/);
+const examples = mdxs.keys().map(mdxs);
+
+## AI2 Header
+
+This component renders a document header that is unified for all AI2 properties.
```
codeviking, line 14:
```suggestion
This component renders a document header that is consistent across AI2 properties.
```

demo/src/pages/Header/page.mdx
```diff
@@ -0,0 +1,28 @@
...
+
+### When To Use
+
+You should put this at the top of the page. It must be present on every page of your application.
```
codeviking, line 18: I think we may only require the `<AI2Banner />`. The `<Header />` is nice too, but I think there's cases where it doesn't make sense to use (like PAWLS).

demo/src/pages/Header/page.mdx
```diff
@@ -0,0 +1,28 @@
...
+
+### Examples
+
+<Examples cols={frontMatter.cols} examples={examples} />
+
+### API
+
+| Property      | Description                                   | Type    | Default |
+| ------------- | --------------------------------------------- | ------- | ------- |
+| alwaysVisible | prevent top banner from hiding when scrolling | boolean | false   |
```
codeviking, line 28:
```suggestion
| alwaysVisible | Prevents the AI2 banner from disappearing when the page is scrolled | boolean | false   |
```

codeviking (commented):
It's cool to see things coming together, and I'm excited to `import { AI2Banner } ...`.

I think we may want to revise the `<Header />` API a bit, so it's more flexible -- let me know what you think.

components/Header.tsx
jonborchardt: ya, layout will be a larger pr later...

components/Header.tsx
jonborchardt: ill give it a shot... but as i recall... this is very specific code to make it work correctly inside a antd.layout.header with a logo

components/Header.tsx
jonborchardt: ill give it a shot

components/Header.tsx
jonborchardt: ill give it a shot, but right now i think the image is clickable...

components/Header.tsx
jonborchardt: looks great

components/Header.tsx
jonborchardt: ya, this change was already planned

demo/src/pages/Header/demo/default.mdx
jonborchardt: k

demo/src/pages/Header/demo/default.mdx
jonborchardt: ill rename... 
im not ready to use the ant columns yet...

components/Header.tsx
codeviking: Is the goal to achieve a specific height? Is it really important that it's *exactly* right, or would it be ok if it's a few pixels off?

demo/src/pages/Header/page.mdx
codeviking: Yea, it is -- but I think sometimes it makes sense for apps to derivate a bit. In this case we wanted to make as much vertical real-estate available for the PDF as possible, so having a large header at the top that takes up real-estate (with nothing but a logo in it), didn't make sense.

components/Header.tsx
jonborchardt: fair enough.. we can make our overall header 1 px taller ;-P

components/Header.tsx
jonborchardt: note, header is capitalized ;-P

codeviking: Haha, yea, I'm trying to embrace that pattern, as it's use widely in `antd`, so I should use it too :).

jonborchardt: so we want a single href that is used on the logo and title, and we dont want to have the logo or title get an underline
```
<Header>
  <Header.Columns>
    <a href="bla"> // <-- optional, a user may not do this
      <Header.Logo>
        <svg>...</svg>
      </Header.Logo>
      <Header.Title>
    </a>
    <Header.Menu>...</Header.Menu>
  </Header.Columns>
</Header>
```
so i dont know about adding the a if href exists

jonborchardt: i could make a titleLogo component...

codeviking: Gotcha. Yea.  That makes sense.

Do we end up with styles coupled to the `a[href]`? That's what I suppose I'm hoping to avoid -- maybe we in this case *do* use a `<Header.Link />`?

codeviking (approved):
Directly we're aligned. There's some cleanup and revisions to the API as noted, ship it once you've got this wired up and feel good about them!

components/Header.tsx
jonborchardt: made inline block, but still needed the margin 0

components/Header.tsx
jonborchardt: done

components/Header.tsx
jonborchardt: this was still needed, feel free to try to remove it :-P

components/Header.tsx
jonborchardt: done

components/Header.tsx
jonborchardt: checked in, see if you like it now

components/Header.tsx
jonborchardt: done

demo/src/pages/Header/demo/default.mdx
jonborchardt: name changed

demo/src/pages/Header/page.mdx
jonborchardt: updated

components/header/Header.tsx
```diff
@@ -109,22 +113,68 @@ const Sticky = styled(LayoutHeader)`
...
+};
+
+const DecorationlessAnchor = styled.a`
+    &&:hover {
+        text-decoration: none;
+    }
+`;
+
+export const Logo = (props: {
+    label?: React.ReactElement<typeof Label>;
```
jonborchardt, line 162: i wonder if this should take ANY reactElement

components/header/index.ts
```diff
@@ -0,0 +1,10 @@
+import { AI2Banner } from './AI2Banner';
+import { Header as HeaderImpl, Columns, Label, Logo } from './Header';
+
+const Header = HeaderImpl as any;
```
jonborchardt, line 4: i could do this in the code, but i think its cleaner in the index, since its just about how we want to export the module

components/Columns.tsx
```diff
@@ -2,14 +2,14 @@ import styled from 'styled-components';
 
 export interface Props {
     count?: number;
-    gridTemplateColumns?: string;
+    columns?: string;
```
codeviking, line 5: 🤘

components/header/Header.tsx
```diff
@@ -109,22 +113,68 @@ const Sticky = styled(LayoutHeader)`
...
+
+const ConditionalAnchorWrapper = (props: { href?: string; children: React.ReactElement }) => {
+    return props.href ? (
+        <DecorationlessAnchor href={props.href} target="_blank" rel="noopener noreferrer">
+            {props.children}
+        </DecorationlessAnchor>
+    ) : (
+        props.children
+    );
+};
```
codeviking, line 153: 👍

components/header/Header.tsx
```diff
@@ -109,22 +113,68 @@ const Sticky = styled(LayoutHeader)`
...
     }
 `;
 
-export const HeaderColumns = styled(Columns)`
-    width: 100%;
-    align-items: center;
-`;
-
-export const HeaderTitle = styled.h5`
+export const Label = styled.h5`
```
codeviking, line 120:
```suggestion
export const AppName = styled.h5`
```

I think `Label` is overloaded. There's an HTML `<label />` element. I think `AppName` makes it clear, we could also do `Name`, but that's a little less clear I think.

components/header/Header.tsx
codeviking: Yea, I like that.

components/header/index.ts
codeviking: Ah, you figured it out. The blessed `any` saves us again.

I think you're right. I'd just add a comment here explaining why the `any` is ok :).

demo/src/AppChrome.tsx
```diff
@@ -60,16 +57,14 @@ export default class AppChrome extends React.PureComponent<RouteComponentProps,
...
-                            <HeaderTitle>{packageJson.version}</HeaderTitle>
-                        </HeaderColumnsWithSpace>
+                        <Header.Columns columns="auto auto 1fr">
+                            <Header.Logo
+                                label={
+                                    <Header.Label>Varnish ({packageJson.version})</Header.Label>
+                                }>
+                                <img src={logoWithText} height={'56px'} alt="Varnish" />
+                            </Header.Logo>
+                        </Header.Columns>
```
codeviking, line 67: This looks awesome!

demo/src/pages/Header/demo/default.mdx
```diff
@@ -0,0 +1,18 @@
...
+
+Default Usage.
+
+```jsx live
+import { Header } from '@allenai/varnish/components';
+
+render(
+  <Header>
+    <Header.Columns columns="auto 1fr">
+      <Header.Logo label={<Header.Label>Title</Header.Label>} />
```
codeviking, line 14: As noted, I'd change `Label` to `AppName` or something like that to avoid confusion with form labels.

codeviking (approved):
LGTM! One suggested name switch, then this is good to go. Nice work!

PR #311, opened 2020-10-15
add component filtering

jonborchardt:


demo/src/AppChrome.tsx
```diff
@@ -97,14 +113,29 @@ export default class AppChrome extends React.PureComponent<RouteComponentProps,
...
+                                            .filter(
+                                                (route) =>
+                                                    !this.state.search.trim() ||
+                                                    route.label
+                                                        .toLowerCase()
+                                                        .includes(
+                                                            this.state.search
+                                                                .trim()
+                                                                .toLowerCase()
+                                                                .toLowerCase()
```
aimichal, line 126:
```suggestion
                                                                .toLowerCase()
```

aimichal: Maybe helpful to name this filter as a function; kinda hard to read to me.

demo/src/AppChrome.tsx
```diff
@@ -73,6 +79,17 @@ export default class AppChrome extends React.PureComponent<RouteComponentProps,
...
+                                value={this.state.search}
+                                placeholder="Filter Components"
+                                onChange={(e) => {
+                                    this.setSearch(e.target.value);
+                                }}
+                                autoFocus
+                                suffix={<FilterOutlined />}
+                                allowClear
+                                bordered={false}
+                            />
```
codeviking, line 92: It'd be cool to make a key binding that auto-focuses this. Like how pressing `T` when viewing a repo on GitHub goes into fuzzy find mode automatically.

I bet users would really like this. Look at doc for `Header`, hit `T`, type `F`, select `Footer`... and so on. It'd allow people to jump really quickly through the docs...

aimichal (approved):

demo/src/AppChrome.tsx
codeviking, line 126: I think you only need one `toLowerCase()` here :)

demo/src/AppChrome.tsx
```diff
@@ -164,29 +195,32 @@ class ImgIcon extends React.PureComponent<{
...
-                    <span>{route.label}</span>
-                </React.Fragment>
-            )}
-        </InternalLink>
+        <Tooltip title={menuCollapsed ? route.label : null}>
+            <InternalLink to={route.path}>
+                {!menuCollapsed ? (
+                    <IconMenuItemColumns>
+                        {!route.icon && route.iconSrc ? <ImgIcon src={route.iconSrc} /> : null}
+                        {route.icon ? route.icon : null}
```
codeviking, line 203: I think it'd be easier to read if these two were collapsed into one. I'd probably make a little helper, or something:

```
getIcon(route)
```

```
function getIcon(route) {
   if (!route.icon) {
       if (route.iconSrc) {
           return <ImgIcon src={route.iconSrc} /> : null;
       }
       return null;
   }
   return route.icon;
}
```

That said I might also revise the data model a bit, so it's all captured in `route.icon`:

```
interface Icon {}
interface ImgIcon extends Icon {
   src: string;
}
interface ComponentIcon extends Icon {
   component: React.ReactComponent;
}
```

Then you can write a function like:

```
function isImgIcon(icon: Icon) icon is ImgIcon {
   return 'src' in icon;
}
```

Then do:

```
{ !route.icon ? null : isImgIcon(route.icon) ? <ImgIcon src={route.icon.src} /> : route.icon }
```

WDYT?

codeviking (approved):
I like it. I left a suggestion, but I don't think it's a hard-stop.

demo/src/AppChrome.tsx
jonborchardt: maybe

jonborchardt: for a double dose of the lowercasing!

demo/src/AppChrome.tsx
jonborchardt: i opted to just remove imgSrc
see the code if you care

PR #312, opened 2020-10-16
add theme mdx

jonborchardt:
this needs a bit of work to cleanup the `typeof Color` crap
but the pr is ready for review


demo/src/App.tsx
```diff
@@ -50,11 +50,11 @@ export default class App extends React.PureComponent {
                 <React.Fragment>
                     {/* <ScrollToTopOnPageChange /> TODO: put this back as soon as
                       @allenai/varnish-react-router is based on this version of varnish */}
-                    <ThemeProvider>
+                    <Theme.Provider>
                         <MDXProvider components={components}>
                             <Route path="/" component={AppChrome} />
                         </MDXProvider>
-                    </ThemeProvider>
+                    </Theme.Provider>
```
codeviking, line 57: I like this new thing we're doing!

demo/src/pages/Colors/ColorValues.tsx
```diff
@@ -157,7 +157,7 @@ export class ExtendedColorValues extends React.PureComponent {
 }
 
 interface PrimaryColorProps {
-    color: Color;
+    color: typeof Color;
```
codeviking, line 160: Isn't each `Color` an instance of a `Color` or something like that? Could we just export the type and use that instead, instead of sprinkling `typeof Color` throughout?

demo/src/pages/Colors/ColorValues.tsx
```diff
@@ -180,7 +180,7 @@ class PrimaryColor extends React.PureComponent<PrimaryColorProps> {
 }
 
 interface ColorGroupProps {
-    colors: Color[];
+    colors: typeof Color[];
```
codeviking, line 183: Is this the `typeof` of `Color[]`, or `(typeof Color)[]`? Maybe parens aren't required, but they might make it more clear.

You could also make a local type, like so:

```
type ColorInstance = typeof Color;
```

That said I'd just import the local type and use it here.

demo/src/pages/Palette/PaletteValues.tsx
```diff
@@ -77,13 +77,13 @@ class ColorGroup extends React.PureComponent<ColorGroupProps> {
                     <BodyMicro>theme.palette.{this.props.keyName}.</BodyMicro>
                 </Namespace>
                 {Object.entries(Theme.palette[this.props.keyName]).map(
-                    ([key, color]: [string, Color]) => {
+                    ([key, color]: [string, any]) => {
```
codeviking, line 80: Why `ant`? Could we use a TypeScript type-guard here instead?

demo/src/pages/Theme/page.mdx
```diff
@@ -0,0 +1,48 @@
+---
+type: Style
+title: Theme
+cols: 1
+iconSrc: ./palette-14px.svg
+---
+
+## Theme
+
+The default Varnish theme, which provides programmatic access to the default values for colors, typography and other properties related to the user interface.
```
codeviking, line 10:
```suggestion
The default Varnish theme provides programmatic access to AI2 branded colors, typography and other properties related to the user interface. Using these values helps ensure your application's look and feel matches that of other AI2 applications.
```

demo/src/pages/Theme/page.mdx
```diff
@@ -0,0 +1,48 @@
...
+import styled from 'styled-components';
+
+const RedBackground = styled.div`
+  background: ${({ theme }) => theme.color.R6};
+`;
+```
+
+## Theme.Provider
+
+Theme Provider for Varnish Theming.
```
codeviking, line 24: Ah, so this isn't the actual context object, but rather an implementation of it, right?

I like the idea of switching to just be called `<VarnishApp>` or `<Varnish>`. Then it can safely act as a generic top-level thing that "activates" Varnish.  This way the user isn't confused by the `Provider` nomenclature, which doesn't make sense unless they're familiar with React.

I also notice that people rarely:

1. Override the theme.
2. Use the `Theme.Consumer`, instead they just import the default theme and use it instead.

So that might influence what we do here...

demo/src/pages/Theme/page.mdx
```diff
@@ -0,0 +1,48 @@
...
+    </Header>
+    <Button type="primary">Click Me!</Button>
+  </Theme.Provider>,
+  mountNode,
+);
+```
+
+### Theme Values
+
+<ThemeValues />
```
codeviking, line 48: We should include a doc for how to use the theme programmatically. I've been asked that 3x times in the last 2 weeks.

I'd suggest we just document how to import and use `Theme.default`, instead of referencing the whole `Theme.Consumer` mechanism. You could include an "Advanced" section where you cover how to override the theme and use `Theme.Consumer` to access the overriden values...this seems like it'd be good to have penciled down so it's easy for a user to do if/when they want to.

theme/index.ts
```diff
@@ -1,2 +1,15 @@
...
-export * from './Theme';
+import { Color, RGB } from './colors';
+import { ThemeProvider } from './ThemeProvider';
+import { Theme as ThemeImpl } from './Theme';
+
+// Adding statics so that we can export a module that is useable as:
+// <Theme>
+// <Theme.Provider>
+// etc
+const Theme = JSON.parse(JSON.stringify(ThemeImpl)); // any is so we can add the random statics below
```
codeviking, line 9: Wait, why are we serializing and then deserializing this? This is somewhat expensive to do, and we'll be doing it every time someone loads a page. There's got to be a better way to do this...

codeviking (changes requested):
I think there's some missing docs and some work to put into thinking about how people access the theme now.

I also don't think we should serialize and then immediately parse the theme. We need to find a better way to do that that isn't a performance hit. Let's chat about why we're doing that.

theme/index.ts
codeviking: We chatted, this is to prevent mutation of the original, we can use `Object.assign` or the merge / spread operator instead: `...`

codeviking (approved):
LGTM; We chatted and:

* We can use `Object.assign({}, ThemeImpl)` to prevent mutating the original reference.
* We're going to dig into why `Theme.Color` isn't what we expect.
* We're going to migrate `Theme.Provider` to a top-level, generic `Varnish` element, but we don't *have* to do that for this to land.

👍

demo/src/AppChrome.tsx
```diff
@@ -60,105 +59,101 @@ export default class AppChrome extends React.PureComponent<RouteComponentProps,
...
-                        <Header.Columns columns="auto auto 1fr">
-                            <Header.Logo
-                                label={
-                                    <Header.AppName>Varnish ({packageJson.version})</Header.AppName>
-                                }>
-                                <img src={logoWithText} height={'56px'} alt="Varnish" />
-                            </Header.Logo>
-                        </Header.Columns>
-                    </Header>
+            <Layout bgcolor="white">
```
jonborchardt, line 62: the only change in this file is removing ln 63

demo/src/pages/Colors/ColorValues.tsx
jonborchardt: fixed, not needing typeof now

demo/src/pages/Theme/page.mdx
jonborchardt: Added VarnishApp

demo/src/pages/Theme/page.mdx
jonborchardt: done... but Theme.default is now just Theme

theme/index.ts
jonborchardt: fixed

README.md
```diff
@@ -83,19 +83,18 @@ Varnish can be used in one of three ways:
...
+        // Wrap your App with the VarnishApp
+        const App = () => (
+            <VarnishApp layoutVariant="app">
+                <MainApp />
+            </VarnishApp>
         );
 
         // Render
-       ReactDOM.render(<VarnishApp />, document.getElementById('root'));
+       ReactDOM.render(<App />, document.getElementById('root'));
```
codeviking, line 97: 👍  This is great.

README.md
```diff
@@ -83,19 +83,18 @@ Varnish can be used in one of three ways:
...
-        const VarnishApp = () => (
-            <ThemeProvider>
-                <App />
-            </ThemeProvider>
+        // Import the parts of Varnish you want. At minimum, you need to import VarnishApp
+        import { VarnishApp } from '@allenai/varnish/components';
+
+        // Wrap your App with the VarnishApp
+        const App = () => (
+            <VarnishApp layoutVariant="app">
```
codeviking, line 91:
```suggestion
            <VarnishApp layoutVariant="left-aligned">
```

I've never understood what `app` means. It's a little too generic for my taste. They're all applications, after all :).

components/VarnishApp.tsx
```diff
@@ -0,0 +1,17 @@
...
+import { Theme } from '../theme';
+
+interface Props {
+    layoutVariant: LayoutVariant;
+    children: React.ReactNode | React.ReactNodeArray;
+}
+
+export const VarnishApp = (props: Props) => (
+    <Theme.Provider>
+        <DefaultLayoutProvider layoutVariant={props.layoutVariant}>
```
codeviking, line 13:
```suggestion
        <DefaultLayoutProvider layoutVariant={props.layoutVariant || "default"}>
```

components/VarnishApp.tsx
```diff
@@ -0,0 +1,17 @@
+import * as React from 'react';
+
+import { LayoutVariant, DefaultLayoutProvider } from '../layout';
+import { Theme } from '../theme';
+
+interface Props {
+    layoutVariant: LayoutVariant;
```
codeviking, line 7:
```suggestion
    layoutVariant?: LayoutVariant;
```

components/VarnishApp.tsx
codeviking: I think we should support them not providing a value -- and do the right thing by default (which is probably a center-aligned layout).

demo/src/pages/Theme/page.mdx
codeviking: Nice !I like that more.

demo/src/pages/Theme/page.mdx
```diff
@@ -0,0 +1,58 @@
...
+
+You can also access the Theme values within javascript via `Theme`.
+
+### Example Usage in javascript
+
+```jsx dark
+import { Theme } from '@allenai/varnish/theme';
+
+const redSix = Theme.color.R6;
+console.log(redSix.hex)
```
codeviking, line 32: If you use red 5 it could be a star wars reference in disguise...

// Red 5 standing by...
const redFiveStandingBy = Theme.color.R5;
```

File: demo/src/pages/VarnishApp/page.mdx [hunk 15]
```diff
@@ -0,0 +1,38 @@
...
+---
+
+## VarnishApp
+
+Outer most container component for all `Varnish` applications.
+
+### When To Use
+
+All `Varnish` applications need to be wrapped in a `VarnishApp` in order to get correct theming and
+context.
```
codeviking, line 15:
```suggestion
All applications need to be wrapped in a `<VarnishApp>`. This ensures Varnish is setup correctly. If you don't do this your application probably won't compile, or if it does things probably won't work quite right.
```

codeviking: I'd leave it more general, we might do more :).

demo/src/pages/VarnishApp/page.mdx
```diff
@@ -0,0 +1,38 @@
...
+
+    // Render
+    ReactDOM.render(<App />, document.getElementById('root'));
+```
+
+### API
+
+| Property      | Description                                                   | Type               | Default   |
+| ------------- | ------------------------------------------------------------- | ------------------ | --------- |
+| layoutVariant | Defines if the header and footer are left or center justified | 'app' \| 'hcenter' | 'hcenter' |
```
codeviking, line 38:
```suggestion
| layoutVariant | Changes the top level layout, which right now only impacts the horizontal alignment of your application. | 'app' \| 'hcenter' | 'hcenter' |
```

theme/index.ts
```diff
@@ -1,2 +1,13 @@
...
+
+// Adding statics so that we can export a module that is useable as:
+// <Theme>
+// <Theme.Provider>
+// etc
+const Theme = Object.assign({}, ThemeImpl) as any; // any is so we can add the random statics below
+Theme.Provider = ThemeProvider;
+
+export { Theme, Color, RGB };
+export type VarnishTheme = typeof ThemeImpl;
```
codeviking, line 13:
```suggestion
import { Theme } from './Theme';

export {
   Provider: ThemeProvider,
   RGB,
   Color,
   ...Theme
}
```

I think you can do ☝️  to avoid the `any` n' all that.

codeviking (approved):
LGTM; Left a few ideas.

README.md
jonborchardt: later (soon) it will go away...

jonborchardt: renamed

components/VarnishApp.tsx
jonborchardt: done

components/VarnishApp.tsx
jonborchardt: done

demo/src/pages/Theme/page.mdx
jonborchardt: nice, done

demo/src/pages/VarnishApp/page.mdx
jonborchardt: done

demo/src/pages/VarnishApp/page.mdx
jonborchardt: done

theme/index.ts
jonborchardt: done

PR #315, opened 2020-10-20
rename context

jonborchardt:


README.md
```diff
@@ -88,7 +88,7 @@ Varnish can be used in one of three ways:
 
         // Wrap your App with the VarnishApp
         const App = () => (
-            <VarnishApp layoutVariant>
+            <VarnishApp appLayoutVariant="center-aligned">
```
codeviking, line 91: Per discussion, I think keeping this simple is nice, since it's user-facing. Maybe just `layout="..."`.

components/AppContext.tsx
```diff
@@ -6,17 +6,17 @@ import React from 'react';
  *  - The "left-aligned" variant is a layout with a left aligned navigation menu. This
  *    is traditionally used for applications with several levels of navigation.
  *
- *  - The "default" variant is a standard, horizontally centered application. This
+ *  - The "center-aligned" variant is a standard, horizontally centered application. This
  *    is a fairly conventional layout used for content  rich websites.
  */
-export type LayoutVariant = 'left-aligned' | 'center-aligned'; // TODO: rename AppLayoutVariant
+export type AppLayoutVariant = 'left-aligned' | 'center-aligned';
```
codeviking, line 12: I'd remove `App` here too.

components/AppContext.tsx
```diff
@@ -25,38 +25,41 @@ interface LayoutSettings {
...
  */
-export const DefaultLayoutProvider = ({ layoutVariant, children }: LayoutProviderProps) => {
+export const DefaultAppLayoutProvider = ({
+    appLayoutVariant,
+    children,
+}: AppContextProviderProps) => {
     const [currentHeaderHeight, setHeaderHeight] = React.useState(0);
     return (
-        <LayoutContext.Provider
+        <AppContext.Provider
```
codeviking, line 56: Per our conversation, `VarnishContext` seems nice too.  It's a bit more specific, since it's not a general context to be used by users per say, but a place where we store stuff we need to share across components.

codeviking (approved):
Some suggested name tweaks, otherwise LGTM.

README.md
jonborchardt: fixed

components/AppContext.tsx
jonborchardt: done

components/AppContext.tsx
jonborchardt: done

PR #321, opened 2020-10-21
add more docs

jonborchardt:


demo/src/ComponentGroups.tsx
```diff
@@ -44,9 +44,14 @@ class ImgIcon extends React.PureComponent<{
     }
 }
 
-const mdGroupTypes = Array.from(
-    new Set<string>(examples.map((ex) => ex.frontMatter.type))
-).sort((a, b) => a.localeCompare(b));
+// predefined order of known types
+const order = ['Ant Design', 'App', 'Patterns', 'Components', 'Extensions'];
```
codeviking, line 48: Right now the documentation doesn't "flow" very well. I think a user might be confused, particularly once they get to Getting Started, as it's a bunch of examples but not really a walkthrough of how to add Varnish to their project.  I myself found myself a bit lost when I started clicking around in the ad-hoc that we got up last night.

Some specifics:

1. I think the welcome should provide a high-level explanation of what Varnish is, and why someone should use it. The important part is to clarify what `antd` is and how `Varnish` works *with* it so that the relationship between these things is clear.

2. Getting started should have more text and less examples. I also think it should be a single column, so it's very clear how to progress through the document. The idea being that it's a guide that the user steps through in a specific order.  It might flow like so:

```
## Getting Started

If you started your application from [Skiff Template](https://github.com/allenai/skiff-template), then you're already using Varnish, as it comes installed by default. You can skip to [usage examples](#usage-examples).

If you didn't, follow these steps to use Varnish:

1. Install `antd` and `@allenai/varnish`
   ```
   yarn add antd @allenai/varnish
   ```

2. Wrap your `<App />` in the `<VarnishApp />` component:
    ```
    // Example here
    ```

## Usage Examples

Now that you have Varnish and `antd` installed, you can start using it to style your application. Below are a few examples:

## Text 

...

## ...

## ...
```

Last, I'd close Getting Started with a bunch of links to the other documents, and explain why people may or may not be interested in them. For instance:

```
- Click [here](./colors) to learn about how to use Varnish colors in custom styles.
- etc
```

None of this needs to be done as part of this PR, but I do think some review and iteration here should be done prior to calling this complete.

demo/src/pages/AppLayout/page.mdx
```diff
@@ -0,0 +1,42 @@
+---
+type: Patterns
+title: App Layout
+cols: 1
+iconSrc: ./header-footer-14px.svg
+---
+
+## Menus
```
codeviking, line 8: Is this title right? It seems unrelated to the content.

demo/src/pages/AppLayout/page.mdx
```diff
@@ -0,0 +1,42 @@
...
+
+To achieve an application layout like those on the
```
codeviking, line 10:
```suggestion
To achieve an application layout like that on the
```

demo/src/pages/AppLayout/page.mdx
```diff
@@ -0,0 +1,42 @@
...
+[Computer Vision Explorer](https://vision-explorer.allenai.org/) and
+[AllenNLP demo](https://demo.allennlp.org/reading-comprehension), you can use this structure.
+
```
codeviking, line 13:
```suggestion
[Computer Vision Explorer](https://vision-explorer.allenai.org/) use the code shown below.

```

demo/src/pages/AppLayout/page.mdx
```diff
@@ -0,0 +1,42 @@
...
+                            )}
+                        />
+                    ))}
+                </Switch>
+            </Content>
+            <Footer />
+        </Layout>
+    </Layout>
+</Layout>
+```
```
codeviking, line 42: Can we include `import` statements in these examples? That way they're copy-pastable.  

Maybe we cut a ticket to do this in a single pass to all examples?

demo/src/pages/Menu/page.mdx
```diff
@@ -0,0 +1,202 @@
...
+type: Patterns
+title: Menu
+cols: 1
+iconSrc: ./links-14px.svg
+---
+
+## Menus
+
+To achieve a menu like those on the [Computer Vision Explorer](https://vision-explorer.allenai.org/)
+and [AllenNLP demo](https://demo.allennlp.org/reading-comprehension), you can follow these steps.
```
codeviking, line 11: I'd drop to a single example here too.

demo/src/pages/Menu/page.mdx
```diff
@@ -0,0 +1,202 @@
...
+
+### InternalLink
+
+First, you need to either use [@allenai/varnish-react-router](https://www.npmjs.com/package/@allenai/varnish-react-router) `<Link>` OR define your own:
```
codeviking, line 15: I think this deserves more explanation. What if they're not using `react-router`?

demo/src/pages/Menu/page.mdx
```diff
@@ -0,0 +1,202 @@
...
+
+```jsx dark
+import styled from 'styled-components';
+import { Link } from 'react-router-dom';
+
+const InternalLink = styled(Link)`
+    ${LinkCSS.default()};
+`;
+```
```
codeviking, line 24: Is this actually necessary? Are we sure that the `<Link />` element doesn't get this style applied by default? Could we add this to `@allenai/varnish-react-router` so they don't have to do this?

codeviking (approved):
I think there's some massaging to be done with respect to the documentation, as right now it feels a little disconnected to me. That said I think this can be handled in the future.

I also think we should probably just put the react-router specific stuff in some sort of advanced section. Users will most likely be using our template, which uses react-router, so trying to get them to understand what applications look like without it just adds details that might confused them.

demo/src/ComponentGroups.tsx
jonborchardt: updated docs a bit, but we do need more

demo/src/pages/AppLayout/page.mdx
jonborchardt: fixed

demo/src/pages/AppLayout/page.mdx
jonborchardt: fixed

demo/src/pages/Menu/page.mdx
jonborchardt: fixed

demo/src/pages/Menu/page.mdx
jonborchardt: fixed

demo/src/pages/Menu/page.mdx
jonborchardt: fixed

